PR #19, opened 2026-09-11
fix(Core): Restore login and natural regeneration

jealous-sound:

New non-GM accounts were disconnected because the Ascension client sends pings every five seconds, while the stock server expects at least 27 seconds. For loopback connections with Ascension compatibility enabled, accept that cadence with one second of jitter tolerance. The overspeed strike limit remains enforced; other connections keep the stock threshold.

Natural health and mana regeneration failed for custom classes because the float-only game-table DBCs were rejected in favor of stock SQL rows. Load their implicit row IDs while preserving explicit-ID formats and SQL overlay precedence, so the client’s coefficients remain available for all 32 classes.

- Native regression harness: 17/17 checks passed, including the installed 3,200-row regeneration tables, actual Player regeneration methods, ping cadence, flood rejection, GM handling and SQL overlays. Running against the previous production code reproduced the loading and ping failures.
- Scoped C++ lint, repository publication checks and `git diff --check` passed.
- RelWithDebInfo worldserver build and local startup passed, with no new startup error lines.
- The user tested the deployed build and confirmed non-GM login and idle health/mana regeneration work. The agent did not launch the game client. Remote sessions retain stock ping timing.

## Data/source dependencies

Uses the existing Ascension game-table DBCs and the five-second ping cadence verified in the local Extensions.dll. No client-file changes are needed.

Applied SQL is unchanged. No credentials, profiles or game archives are included.

AI assistance: Codex generated the code, regression tests and PR description; the user performed in-game acceptance.

### DIFF
--- a/.github/workflows/quality.yml
+++ b/.github/workflows/quality.yml
@@ -17,5 +17,6 @@ jobs:
         with:
           python-version: '3.12'
       - run: python -B apps/codestyle/tests/test_scoped_lint.py
+      - run: python -B modules/mod-ascension-compat/tests/client_compat/run.py
       - run: python -B tools/check_repository.py
       - run: git diff --check HEAD^ HEAD
--- a/modules/mod-ascension-compat/README.md
+++ b/modules/mod-ascension-compat/README.md
@@ -20,3 +20,28 @@ directory, preserve applied SQL hashes, and follow
 `.agents/docs/systems/ascension-local-deployment.md` in the server checkout.
 Do not run historical installers or clear client caches to compensate for a
 protocol, data or process problem.
+
+## Login and natural regeneration
+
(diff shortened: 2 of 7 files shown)


andyger1994:
Hello,

I am setting up your AzerothCore CoA repository, but I am missing the compatible Ascension/CoA client files.

The server starts successfully, but with a standard WotLK 3.3.5a client it behaves like a normal AzerothCore server and reports many missing custom spells and DBC entries.

Would you be able to publish the compatible client files, or provide a legal download/source and instructions for obtaining and extracting them? If redistribution is not possible, checksums and the exact required client version would also be very helpful.

Thank you for your work on the project.

e1337ebird:
Thank you for this project. I don't see any documented place where this question is asked, but what version of the COA client is compatible? Again, thank you for this project

PR #67, opened 2026-09-13
feat(Core): add the core hooks mod-playerbots requires

steviecraycray:
## What this is

`mod-playerbots` is not a self-contained module. Upstream it ships next to a **patched** AzerothCore ([mod-playerbots/azerothcore-wotlk](https://github.com/mod-playerbots/azerothcore-wotlk), branch `Playerbot`) — drop the module into a stock core and it compiles, registers, and then does nothing, because the hooks it relies on do not exist.

This PR brings those core patches into this fork, adapted to CoA. It is the prerequisite for running playerbots here; it does not add any bot logic itself.

## Safety of the change

Everything is **additive**. No existing signature changes, no existing behaviour changes.

`MOD_PLAYERBOTS` is only defined when `modules/mod-playerbots` is actually present in the tree, so a build without the module is byte-for-byte the core you have today. The same guard now also controls whether `modules` links against `mysql`.

## What is in it

**Hook families**

| | |
|---|---|
| `DatabaseScript` | six new hooks (`OnDatabasesLoading` / `Closing` / `KeepAlive`, `GetDBRevision`, `SelectIndexLogout`, `WarnAboutSyncQueries`) plus their five call sites. Without these the module's database is never opened. |
| `PlayerbotScript` | new registry type |
| `PlayerScript` | `OnPlayerAfterUpdate` — the tick the bot engine runs on |
| `ServerScript` | `OnPacketReceived` |
| new files | `PlayerbotsDatabase.cpp/.h`, `PlayerbotsScript.cpp` |
| `DBUpdater` | resolves its source directory per database, so the module's SQL under `modules/mod-playerbots/data/sql` is found. Without this the playerbots updates silently never run and `BisListMgr::LoadAll` aborts on a missing table. |

**Core API bots need because they have no client session**

- `Guild`: public session-free variants of `HasRankRight`, `GetRankRights`, `MemberHasTabRights`, `HandleSetEmblem`, `HandleSetRankInfo`
- `WorldSession::IsBot`, `GetPacketQueue`; `LoginQueryHolder` moved from `CharacterHandler.cpp` into the header so modules can derive from it (the definition of `Initialize()` stays in the .cpp)
- `World::AddQueryHolderCallback`, `Player::BotCanUseItem`, `Group::GetRolls`, battleground state accessors for AV/AB/EY/IC, a `CharacterCreateInfo` constructor taking appearance fields directly, `ChatHandler::BuildChatPacket` without a `WorldObject`, `Item::CreateItem(temp)`, `Graveyard::GetClosestGraveyard` by coordinates
- `CharSections.dbc` and `EmotesTextSound.dbc` loaded

## Four places where I deliberately did *not* follow the upstream patch

These are the ones worth your eyes — each is a spot where the playerbot fork's version would have damaged CoA:

1. **Graveyard lookup** keeps `IsClass(..., CLASS_CONTEXT_GRAVEYARD)`. The fork uses `getClass() == CLASS_DEATH_KNIGHT`, which throws away the CoA class logic.
2. **`SpellAuraDefines.h` untouched.** The fork's patch would have deleted 76 lines of CoA aura extensions.
3. **`GetAreaEntryByAreaID`** builds its own id mapping on first access, because this core keeps no `AreaFlag` table.
4. **`MemberHasTabRights`** repeats the logic instead of forwarding, because `_MemberHasTabRights` is defined inline in the .cpp here.

## Also

- `deps/boost` gains the `thread` component — under MSVC, Boost requests the library itself via `#pragma comment(lib, ...)`, so without listing it CMake never looks for it and the link fails.
- `ScriptMgr` static_assert counters move to 49/35 for the new registry type.

## How it was verified

Built and started on Windows / MSVC (VS2019, `RelWithDebInfo`) with both `mod-ascension-compat` and `mod-playerbots` enabled, against a CoA 12340 client. Bots log in, move, take targets and cast. The repository checks in `quality.yml` pass locally.

What is **not** verified: nothing here is tested on Linux/GCC, and I have not run a build with the module absent on CI — only locally.


steviecraycray:
Schliesse den hier erst mal. Der Branch ist inzwischen 13 Commits hinter main, und die Kernhooks sind nur ein Teil dessen, was wir inzwischen haben - dazu kommen drei Serverabstuerze im Compat-Modul und ein Buildfehler im neuen Testharness. Wir arbeiten das lokal sauber durch und melden uns mit einem vollstaendigen PR, sobald die Klassen auf Stufe 60 wirklich spielbar sind. Danke fuers Offenhalten.

Issue #193, opened 2026-09-14
Can't change xp rate


jealous-sound:
Rate.XP.Kill           = 1
Rate.XP.Quest          = 1
Rate.XP.Explore        = 1
Rate.XP.BattlegroundKill = 1

Changing those doesn't seem to do anything

lostmind84:
We could not reproduce this: on the current fork code, `Rate.XP.Kill` and `Rate.XP.Quest` set in `worldserver.conf` do change the experience players receive.

## How we tested

We added an XP scenario to the gameplay test harness (`apps/coa-gameplay-test`), on branch
[`feat/gameplay-test-xp-rates`](https://github.com/lostmind84/azerothcore-wotlk-coa/tree/feat/gameplay-test-xp-rates)
of lostmind84/azerothcore-wotlk-coa.

- The scenario ([`scenarios/xp-rates.json`](https://github.com/lostmind84/azerothcore-wotlk-coa/blob/feat/gameplay-test-xp-rates/apps/coa-gameplay-test/scenarios/xp-rates.json)) runs twice, each time in its own disposable worldserver with copied databases: once with both rates at 1, once with both at 10.
- The rates are written into the generated `worldserver.conf`, so the server loads them through its normal startup configuration, and the server reports back the values it actually read.
- A level 15 character kills a creature (entry 36) and is then rewarded quest 46 (Bounty on Murlocs) through console commands. The XP gained from each is measured.

Results on a Linux Docker install:

| | Rates at 1 | Rates at 10 | Ratio |
|---|---|---|---|
| Creature kill | 75 XP | 750 XP | 10 |
| Quest reward | 1050 XP | 10500 XP | 10 |

Not covered: `Rate.XP.Explore` and `Rate.XP.BattlegroundKill*`.

## Testing it yourself

You need a working CoA server (MySQL databases, client data, module configs). On Linux with the Docker stack, from a normal clone (not a git worktree, since the Docker build mounts `.git`):

```bash
git remote add lostmind84 https://github.com/lostmind84/azerothcore-wotlk-coa.git
git fetch lostmind84 feat/gameplay-test-xp-rates
git switch -c feat/gameplay-test-xp-rates lostmind84/feat/gameplay-test-xp-rates

mkdir -p .cache/coa-gameplay-tests
# The xp-rates tag keeps your live worldserver image (tag master) untouched.
DOCKER_IMAGE_TAG=xp-rates docker compose build ac-worldserver
[...]

PR #278, opened 2026-09-15
feat(CoA): run gameplay tests on Linux via Docker

lostmind84:
## Summary
- The gameplay test runner (`apps/coa-gameplay-test`) only worked on Windows. This makes it run on Linux against a Docker Compose install, with no C++ or SQL change.
- `AC_*` environment variables override worldserver config values, and the worldserver image sets `AC_UPDATES_ENABLE_DATABASES=0`. The test worldserver no longer inherits `AC_*` variables that would override generated harness values.
- Linux worldservers read module configs from `CONF_DIR/modules`, fixed at build time. New `--server-modules-dir` option, required outside Windows; existing files there are never replaced.
- Database credentials and the generated config now stay in a private temp directory, never in the results directory, and SIGTERM runs the normal cleanup.
- New `ac-gameplay-test` Compose service (`apps/coa-gameplay-test/docker/`, profile `tests`). It shares the `ac-database` network namespace, so MySQL stays `127.0.0.1` and the local database checks are unchanged. Decision recorded in `docs/adr/0001-linux-docker-gameplay-test-service.md`.

Groundwork for XP rate tests related to #193.

## Test plan
- [x] `python apps/coa-gameplay-test/test_runner.py`: 25 tests pass
- [x] Arch Linux, local Docker images: `frostbolt.json` passed through the Compose service (exit 0, 8/8 assertions), test schemas dropped, no credential files in results
- [ ] Windows: run a scenario with the existing README command to check nothing regressed

AI assistance: prepared with Claude Code (Claude Opus 5).


lostmind84:
Ran the service for real on Linux (Arch host, Docker Compose, worldserver image built from this branch). Both runs passed:

| Run | Scenario | Result | World cache | Prepare | Server | Total |
|---|---|---|---|---|---|---|
| `1eaad147ddc4` | `frostbolt.json` | passed, 8 assertions | `created`, retained | 271.9 s | 33.4 s | 318.1 s |
| `4848fe23b5f5` | `starcaller-bright-moon-lunar-phase.json` (#3960) | passed, 6 assertions | `reused` | 25.9 s | 56.1 s | 92.4 s |

```
Run 1eaad147ddc4: /results/20260918T074237765209443Z
Checking source world data for changes...
Preparing reusable world database...
Worldserver harness is ready; executing scenario...
PASSED: Frostbolt: learned spell, damage, slow, mana and expiry
```

What this confirms, beyond the scenarios passing:

- World-cache reuse works across runs ("Reusing isolated world database; full copy skipped"), and the cache survives the disposable container because the metadata now lives on the host mount — one of the changes in 46a9a1d4.
- Cleanup is correct: only the retained `coa_test_<id>_world` schema is left; both disposable auth/characters schemas are gone.
- The live worldserver kept running untouched through both runs, so the service can be used against a server that is already up.
- It works inside a non-default Compose project (a separate stack with its own project name, image tag, env file and volume paths), so `DOCKER_IMAGE_TAG`, `DOCKER_AC_ENV_FILE` and `DOCKER_VOL_*` are honored.

Two caveats, stated exactly:

1. I tested 46a9a1d4 with current `origin/main` merged into it, because my preflight refuses a checkout that is behind `origin/main`. The branch as it stands was not the tested tree; merging `main` (or rebasing) before merge would make the tested state the shipped state.
2. The server would not start at all until I created `creature_display_preset` by hand. Since #3983, `AscensionCreaturePresetMgr::LoadFromDB` selects from that table, no migration in the repository creates it, and the failed query aborts startup on any realm whose world database predates the feature:

```
[1146] Table 'acore_world.creature_display_preset' doesn't exist
Your database structure is not up to date.
```

I reproduced that twice on a database freshly seeded from the repack dump with all 923 shipped updates applied. #3995 and #3999 both propose the missing table with identical columns; I applied that schema locally, without committing anything. Whichever of the two you take, it is a prerequisite for anyone rebuilding a realm from the public dump — and for this harness, since the runner clones the same world database.

jealous-sound (approved):

Issue #110, opened 2026-09-14
Primalist Auto Attacks generate no rage


Kalandorno:
### Core revision and client profile

Ascension client iwth replaced Oatch-B and DLL

### Reproduction, expected result and actual result

Auto attack, no rage generates

lostmind84:
Confirmed, same cause as #213 (details in #340).

PR #397, opened 2026-09-16
fix(CoA): let the Witch Hunter gain Rage from combat

### LINKED ISSUE #213: Witch Hunter Rage

#### Issue
Witch Hunter does not gain rage on melee or taking damage. Rage is only generated by using mana abilities.
Yes

### Classification
Category: 9
Severity: 3

Client: 3.3.5 (12340)

### Server context
Class ID: 15
Level: 8
Map ID: 1
Position: -1241.18, -5462.19, 5.27141
Core revision: feb155c82c75

jealous-sound:

"Witch Hunter does not gain rage on melee or taking damage. Rage is only generated by using mana
abilities."

Both stock Rage grants live in `Unit::DealDamage` and are gated the same way:

```cpp
// Rage from Damage made (only from direct weapon damage)
if (attacker && cleanDamage && damagetype == DIRECT_DAMAGE && attacker != victim && attacker->HasActivePowerType(POWER_RAGE))
...
// Rage from damage received
if (attacker != victim && victim->HasActivePowerType(POWER_RAGE))
```

`Player::HasActivePowerType` asks the player scripts (`OnPlayerHasActivePowerType`) and otherwise falls
back to `getPowerType() == power`. `ChrClasses.dbc` gives class 15 (Witch Hunter) **Mana** as its display
power and no script answered for it, so both grants were skipped — even though the class genuinely runs
on Rage: 216 records in its spell family (21) carry a Rage cost, Quickdraw (806847) among them, and the
module already spends Rage in `AscensionWitchHunterAbilities.cpp` and tops it up per ability.

This adds a small `PlayerScript` answering the hook for the class, mirroring `starcaller_resources`,
which already answers for Mana and Energy.

Ability damage is not double-counted: the damage-dealt path only fires for `DIRECT_DAMAGE` with a melee
`attackType`, i.e. auto attacks, so the module's explicit `RewardRage` call for Heartseeking still
stands alone. The only other consumer of the flag for Rage is `Player::ResetAllPowers`, which now zeroes
the bar on reset as it does for other Rage users.

- `python apps/codestyle/codestyle-cpp.py --files modules/mod-ascension-compat/src/AscensionWitchHunterCompletion.cpp` — passed.
- Evidence: the two gates in `Unit::DealDamage`, `Player::HasActivePowerType`, the `ChrClasses.dbc`
  record for class 15, and a scan of the shipped `Spell.dbc` for family-21 records with `PowerType = RAGE`.
- **Not built and not gameplay-tested** — source-only run without build authorization; the Rage bar has
  not been watched in game.
- Related but untouched: #110 reports the same symptom for the Primalist (class 31, also a Mana display
  power). I found no Rage costs or Rage handling for that class in the module, so it needs its own
  evidence rather than being folded in here.

## Data/source dependencies

No SQL in this change; applied SQL is unchanged. No credentials, profiles or game archives are included.

Fixes #213


jealous-sound:
Review finding confirmed and fixed separately in #406.

A fully absorbed swing does reach `Unit::DealDamage` (`DealMeleeDamage` only skips when damage, absorb and resist are all zero), `rage_damage` already carries the absorbed amount, so the melee branch pays the attacker and the `if (!damage)` block then paid again. That block also had no damagetype/attack-type gate, so fully absorbed spells and ranged attacks paid attacker Rage too.

This corrects my description above: the "ability damage is unaffected" note held for the melee branch only, not for the absorbed block.

#406 keeps only the victim's reward in that block, so an eligible melee swing is paid once by the melee branch. It touches `Unit.cpp` only, so it does not conflict with this PR — please merge it first.

PR #398, opened 2026-09-16
fix(CoA): let the Primalist gain Rage from combat

### LINKED ISSUE #110: Primalist Auto Attacks generate no rage

jealous-sound:

Primalist auto attacks generated no Rage. Both stock Rage grants in `Unit::DealDamage` — from melee
damage dealt and from damage received — are gated on `HasActivePowerType(POWER_RAGE)`, which asks the
player scripts (`OnPlayerHasActivePowerType`) and otherwise falls back to `getPowerType() == power`.

`ChrClasses.dbc` gives class 31 (Primalist) **Mana** as its display power and nothing answered the hook
for it, so neither grant ever ran — even though the class runs on Rage: 123 records in its spell family
(37) carry a Rage cost, Seismic Smash (40 Rage) among them.

The fix adds a small `PlayerScript` answering the hook for the class, mirroring `starcaller_resources`
(Mana/Energy) and the matching Witch Hunter change in #213.

The only other consumer of the flag for Rage is `Player::ResetAllPowers`, which now zeroes the bar on
reset as it does for every other Rage user.

- `python apps/codestyle/codestyle-cpp.py --files modules/mod-ascension-compat/src/AscensionPrimalistSecondary.cpp` — passed.
- Evidence: the two gates in `Unit::DealDamage`, `Player::HasActivePowerType`, the `ChrClasses.dbc`
  record for class 31, and a scan of the shipped `Spell.dbc` for family-37 records with `PowerType = RAGE`.
- **Not built and not gameplay-tested** — source-only run without build authorization; the Rage bar has
  not been watched in game.
- The report does not say whether Rage should build only in a particular form or spec. This restores the
  stock behaviour for the whole class, which is what a Rage-costing ability set needs to be usable; if
  Ascension gated it more narrowly, that needs its own evidence.

## Data/source dependencies

No SQL in this change; applied SQL is unchanged. No credentials, profiles or game archives are included.

Fixes #110


jealous-sound:
Review finding confirmed and fixed separately in #406.

A fully absorbed swing does reach `Unit::DealDamage` (`DealMeleeDamage` only skips when damage, absorb and resist are all zero), `rage_damage` already carries the absorbed amount, so the melee branch pays the attacker and the `if (!damage)` block then paid again. That block also had no damagetype/attack-type gate, so fully absorbed spells and ranged attacks paid attacker Rage too.

This corrects my description above: the "ability damage is unaffected" note held for the melee branch only, not for the absorbed block.

#406 keeps only the victim's reward in that block, so an eligible melee swing is paid once by the melee branch. It touches `Unit.cpp` only, so it does not conflict with this PR — please merge it first.

PR #431, opened 2026-09-16
feat(Module/DB): Worldforged items - restoration

bozo-1:

On CoA a Worldforged item is picked up off the ground: a pouch, a bucket, a pile of bones, a packet - a world object **named after the base item it holds**, openable **once per character**. The database has the items and the object templates, but **no spawns for those objects**, so there is no way to get one.

This restores that layer: **1,555 pickups, 1,510 of them placed**, each holding exactly the base item its own name says it holds.

* `data/sql/db-world/` - the pickups: templates and loot rows written with `REPLACE INTO` (keyed on `entry`, and on `Entry, Item`, every column listed), and each block of spawns preceded by a `DELETE` on its own `guid` range, so re-applying is safe. Undo in `data/sql/manual/`. Templates are the client's own gameobjectcache captures, each loot row is the realm catalog's own row (`gameobject_loot`, chance 100), positions are the community's observed object coordinates.
* `data/sql/db-characters/` - `character_worldforged_loot (guid, spawn_id, entry, looted_at)`, keyed on the spawn id, never the runtime object GUID (this core hands out map-local generated GUIDs, which are not stable across grid reloads).
* `src/WorldforgedPickups.cpp` - the rule the data cannot express: once opened, a pickup is inert **for that character alone** (no sparkle, not selectable) while every other character still sees it sparkling. The item is refused at the award itself (`OnAllowedForPlayerLootCheck`), the use is refused outright (`GossipHello`), a spent loot left open is cleaned up (`OnStateChanged`), and the awarded item plus its ledger row commit in one character-database transaction. The ledger is read in `Player::LoadFromDB` rather than at login, or a relog makes a spent pickup look lootable again.
* `GameObjectAI::BuildClientFlags` - one small generic core hook (`GameObjectAI.h`, `GameObject.cpp`): a gameobject AI can give a single viewer its own `GAMEOBJECT_DYNAMIC` / `GAMEOBJECT_FLAGS`.

- [x] Core
- [x] Scripts/module
- [x] Database

Currently running and played on a repack.

* Boot: `>> Loaded 98138 Gameobjects` against `SELECT COUNT(*) FROM acore_world.gameobject` = 98,138, so nothing was skipped for a display id; 3,784 C++ scripts; no `invalid displayId`, no invalid-rotation warning, and no `Script named 'worldforged_pickup' is assigned in the database, but has no code!`
* Data: 1,555 templates / 1,555 loot rows / 1,510 spawns; every spawn's template resolves to a loot row; every loot row is at chance 100; no looted item missing from `item_template`. The migration reproduces those rows byte-identical on a re-apply.
* Lint: both migrations pass `apps/codestyle/codestyle-sql.py`. The only file of the three that still trips its rule against deleting `gameobject_template` rows is the manual undo, which deletes them on purpose and says so.
* In game: `.go xyz -8769.1 -174.1 83.9` (Wax Stained Bag, 185 yd from Stormwind) sparkles, hands over its own base item once, and is then inert for that character - across a relog too - while another character still sees it sparkling.

**No respawn timer is involved**, so none is invented: the spawns carry `spawntimesecs = 0`. `GameObject::LoadGameObjectFromDB` copies it into `m_respawnDelayTime`, which is only read when something despawns the object - and for a chest with `consumable = 0` the deactivation branch sets `GO_READY` and returns before any respawn can be scheduled, which is exactly the re-roll the next character needs. 0 is also what `ObjectMgr::LoadGameobjects` accepts silently: it only reports a zero spawn time as an error when `IsDespawnAtAction()` is true, and for a chest that is `chest.consumable`.

Limits:

* **Facing is not recoverable.** No available source holds the realm's orientation for these objects - not the client gameobject cache, not the realm's catalog export (its `gameobject_spawn` table is empty and carries no orientation column), not the community position dumps, not the archive atlas. Each spawn therefore carries a stable per-object facing with the matching unit quaternion (`rotation0 = rotation1 = 0`, `rotation2 = sin(a/2)`, `rotation3 = cos(a/2)`, the convention the rest of `gameobject` uses). Stable, so a re-import reproduces the same world. The captures' zero quaternion is not unit length, which the core rejects row by row.
* **Positions are one observation per object**, the honest maximum - the catalog carries no spawns at all. They are the object's own coordinates, not the looter's: for 2,954 matched loot records the distance from the loot position to the observed position equals the recorded distance to the nearest object of that entry (median difference 0.0 yd, 89% within 1.5 yd), and the 13 placements more than 10 yd out are re-anchored to it. Heights are the observation's own, corroborated by the dump's stored object height (175 of the 182 spawns sitting >15 yd below this repack's terrain grid, all genuine interiors, agree within 1.5 yd).
* **The loot interaction itself has no automated test.** The repository's gameplay harness can drive spells, items and talents, but it has no step that takes an item out of a world object, so the award path is verified by code path: `Player::StoreLootItem` asks `LootItem::AllowedForPlayer` for each slot before awarding, and that is where the pickup answers for a character who already looted it.
* The item is per pickup, not a drop table: each pickup holds the one item it is named for, which is what the community data proves.

## Data/source dependencies

* **The client's gameobject display table.** A spawn whose `displayId` is missing from the server's `Data/dbc/GameObjectDisplayInfo.dbc` is dropped at load (`... has an invalid displayId (87226), not loaded.`). A stock table holds 3,792 rows; CoA's client ships 120,871, and the restored pickups use 1,009 ids from it - so against the stock table **1,489 of the 1,510 spawns never reach the world** and the world stays empty however good the SQL is (it also silently drops ~1,222 unrelated stock objects). The script merges the client's rows into the server's without touching rows it already has, backs the file up and verifies the result:

  ```bash
  pip install mpyq
  python modules/mod-worldforged-pickups/tools/install_gameobject_display_info.py --client "C:/path/to/CoA/client"
  ```

* With `Updates.EnableDatabases = 7` (all three databases) the core applies both SQL files by itself; use `6` for the characters and world databases only - `1` is the auth database alone and would leave both migrations unapplied. Where the updater is off, apply them by hand.
* Sources: the client's own gameobjectcache captures, the realm's catalog export, and the community's observed object positions. No SQL that this PR does not add was changed, and no credentials, profiles or game archives are included.

## Undo

```sql
SOURCE data/sql/manual/worldforged-pickups-revert.sql;
DROP TABLE IF EXISTS acore_characters.character_worldforged_loot;
```

and restore `Data/dbc/backup/GameObjectDisplayInfo.dbc.bak-<stamp>`.


jealous-sound:
1. **[P1] Enforce eligibility when taking the item.** [WorldforgedPickups.cpp:258](https://github.com/jealous-sound/azerothcore-wotlk-coa/blob/8e1ab65e98cd16beb1c661c7cb7e3a60180a0e0d/modules/mod-worldforged-pickups/src/WorldforgedPickups.cpp#L258-L268)

   A player can retain their loot session after claiming a pickup. When another player regenerates its shared loot, the first player can request that slot again (after disposing of the first item if its unique limit would otherwise block the award). Neither `StoreLootItem` nor `AllowedForPlayer` checks the pickup ledger, and this callback runs **after** awarding the item. Add an eligibility check before the award, such as through `OnAllowedForPlayerLootCheck`.

2. **[P2] Save the item and pickup history atomically.** [WorldforgedPickups.cpp:133](https://github.com/jealous-sound/azerothcore-wotlk-coa/blob/8e1ab65e98cd16beb1c661c7cb7e3a60180a0e0d/modules/mod-worldforged-pickups/src/WorldforgedPickups.cpp#L133-L135)

   The ledger insert commits independently while the awarded item awaits inventory saving. A crash between those writes permanently marks the pickup as claimed without preserving its item. Persist both in the same character-database transaction.

3. **[P2] Correct the documented database-update setting.** [README.md:69](https://github.com/jealous-sound/azerothcore-wotlk-coa/blob/8e1ab65e98cd16beb1c661c7cb7e3a60180a0e0d/modules/mod-worldforged-pickups/README.md#L69-L73)

   `Updates.EnableDatabases = 1` enables only auth updates. Following these instructions leaves both module migrations unapplied. Use `7` for all databases, or `6` for characters and world.

Reviewed `8e1ab65e98cd16beb1c661c7cb7e3a60180a0e0d`. C++ lint passed; SQL lint failed. No build or gameplay tests were run.

bozo-1:
All four points are addressed in `75c3d97ab` - three commits on top of the reviewed `8e1ab65e9`, plus a merge of `main` to clear the conflict.

**P1 - eligibility at the award.** The pickup now implements `OnAllowedForPlayerLootCheck`, which `LootItem::AllowedForPlayer` consults (`LootMgr.cpp:465`), and `Player::StoreLootItem` calls that for each slot *before* `StoreNewItem` (`Player.cpp:14189`). A character who already looted this pickup is refused the item itself, whatever the request is - click, auto-store or the group window. One thing worth knowing while reviewing: the hook reads backwards. `ScriptMgrMacros.h`'s `CALL_ENABLED_BOOLEAN_HOOKS` returns false as soon as a script returns true, and `AllowedForPlayer` drops the item on that false, so returning true is what withholds it. The class carries that note so nobody "fixes" it into a no-op later.

The refusals are four now, each answering a different question: what this viewer is shown (`BuildClientFlags`), whether the use opens anything (`GossipHello`; `GameObject::Use` returns before `SendLoot`), whether the item may be handed over (this hook), and cleaning up a spent loot left behind (`OnStateChanged`).

**P2 - one transaction.** `WorldforgedLootStore::Claim` writes the awarded item row, its `character_inventory` placement and the ledger row in a single `CharacterDatabaseTransaction` - the same two item writes `Player::_SaveInventory` would have made at the next save, moved to the moment of the award, plus the ledger insert. A failed write now leaves the pickup unclaimed instead of spent without its item.

**P3 - the setting.** Corrected to `Updates.EnableDatabases = 7`, with `6` noted for the characters and world databases, and `1` called out as the auth database alone.

**SQL lint.** Both migrations now pass `apps/codestyle/codestyle-sql.py` clean. The world file was deleting its 1,555 `gameobject_template` rows before inserting them, and only the first block of INSERTs had a DELETE in front of it. Templates and loot rows are `REPLACE INTO` now - keyed on `entry`, and on `Entry, Item`, with every column listed - and each block of spawns is preceded by a DELETE on its own `guid` range. I applied the rewritten file to a live copy of the restored world database: 1,555 templates, 1,555 loot rows and 1,510 spawns come out byte-identical, and identical again on a second apply. The one file that still trips that rule is the manual undo (`data/sql/manual/worldforged-pickups-revert.sql`), because deleting those rows is what an undo is for; it stays out of `db-world/` and its header says why.

**Testing.** Built with MSVC and booted on the repack: `>> Loaded 98138 Gameobjects` (equal to the count in the table), 3,784 C++ scripts, no `invalid displayId`, no invalid-rotation warnings, no ERROR lines. I could not drive the loot itself - the repository's gameplay harness drives spells, items and talents, but has no step that takes an item out of a world object - so the award path is verified by code path: the check sits in `StoreLootItem` before the item is created, and the hook is the only new thing that can answer it. That limit is now stated in the PR body rather than left implicit.

PR #1164, opened 2026-09-16
fix(Core): start the global cooldown for cast-while-casting spells

### LINKED ISSUE #391: Inner Demon Doesn't Consume Felfury

#### Spell
 Inner Demon [ID: 804216]

#### Issue
Inner Demon doesn't consume Felfury, nor triggers global cd, so it can be spammed, which is pretty ffunny
Yes
#### Exact Location
Sunstrider Isle
#### What were you doing
Using Inner Demon
#### Expected Outcome
It would consume all my Felfury
#### Actual Outcome
It didn't consume any Felfury
#### Steps to Reproduce
Step 1: Use Inner Demon

### Classification
Category: 0
Severity: 3

Client: 3.3.5 (12340)

### Server context
Class ID: 14
Level: 4
Map ID: 530
Position: 10374.8, -6210.87, 29.8138
Core revision: a8b28faac98d+

lostmind84:

#391: "Inner Demon doesn't consume Felfury, nor triggers global cd, so it can be spammed."

The Felfury half is already fixed by #340; the report was filed against `a8b28faac98d+`, which predates it. The global cooldown half is real and has two causes, both needed for the fix:

1. **Core.** `Spell::Spell` adds `TRIGGERED_IGNORE_CAST_IN_PROGRESS | TRIGGERED_CAST_DIRECTLY` to every spell carrying `SPELL_ATTR4_ALLOW_CAST_WHILE_CASTING`. In `Spell::prepare`, `TRIGGERED_CAST_DIRECTLY` selects the direct `cast(true)` branch, which skips the whole `else` branch — `TriggerGlobalCooldown` included. Such a spell never started a global cooldown, whatever its `StartRecoveryTime`. The fix starts it on the direct branch as well, before `cast()`, which may finish the spell. Real triggered casts carry `TRIGGERED_IGNORE_GCD` through `TRIGGERED_FULL_MASK` and keep skipping it.
2. **Data.** Inner Demon (804216) has `StartRecoveryCategory 0` and `StartRecoveryTime 0`, so even a fixed prepare path starts nothing, and category 0 is never populated by other spells either. `AscensionFelswornContracts` now gives it category 133 for 1000 ms, matching Ruin, Felwrath, Sunder, Twin Slice and Fel Fireball. It is a melee-class ability (`SPELL_ATTR0_IS_ABILITY`), and 231 of the Felsworn family's spells use 1000 ms against 26 at 1500 ms; this value is a judgement call, not read from a source of truth.

The client reads its own `Spell.dbc`, so `apps/coa-spells/inner_demon_gcd.py` writes the same two fields into a copied client record, following the other `apps/coa-spells` generators. Without it the client does not grey the button and keeps sending casts the server refuses with `NOT_READY`.

**Behaviour change beyond Inner Demon — please confirm it is intended.** The core fix restores the global cooldown the spell data asks for on every spell with the attribute. Among class-learned player spells, seven now get one they never had in game:

| Spell | Class | New global cooldown |
| --- | --- | --- |
| Undead: Assault / Pacify / Protect (500982, 500983, 500985) | Necromancer | 3000 ms, shared in category 222 |
| Sanity Tap (802575) | class 25 | 1500 ms in the common category 133 |
| Vault (500085) | Witch Hunter | 1500 ms in category 25 |
| Build: Sentry Turret (500239) | Tinker | 1500 ms in category 100 |
| Lunar Eclipse (800386, 800429) | Starcaller | 1000 ms in category 100 |

Lunar Eclipse 800386 and Necromancer 500991 are routed to the direct branch explicitly in `Spell::prepare` (46bf8121e); 500991 has no `StartRecoveryTime` and is unaffected.

- Built and run on this exact commit (`eef6f2f73`, on top of current `main`); all shipped SQL updates applied.
- `python3 apps/codestyle/codestyle-cpp.py --files` on every changed C++ file — passed.
- ConquestOfAzerothGhost e2e (mattparkes/ConquestOfAzerothGhost, branch `test/coa-inner-demon-global-cooldown`), run with `go test -p 1`:
  - `e2e/classes/felsworn/felfury`: 1 of 8 rapid Inner Demon casts lands and the other 7 are refused with `NOT_READY`; Inner Demon is refused during Vengeful Pact's cooldown (control: Vengeful Pact refuses its own recast); Felfury is fully consumed with 2, 3 and 6 points, at level 4, when generated by Fel Fireball, on a recast while Inner Demon is up, and when cast during another cast, which still completes.
    - Review follow-up: start Fel Fireball, wait for its own cooldown to end, cast Inner Demon, cancel Fel Fireball, then try Vengeful Pact. Accepted 3/3 on `0eeb57e`, refused with `NOT_READY` 3/3 on `eef6f2f`.
- `e2e/classes/reaper/soulinfusion`, `e2e/classes/stormbringer/static`, `e2e/coa/rage`: pass.
  - `e2e/classes/chronomancer/time`: fails on the existing Aeon mutual-exclusivity bugs, unrelated to this change.
- In game (Linux client under Proton): the server refuses Inner Demon during its own and Vengeful Pact's cooldown. With the client `Spell.dbc` produced by `inner_demon_gcd.py` and packed into a copy of `patch-T.MPQ`, the cooldown sweep shows on Inner Demon and the other action buttons, Vengeful Pact greys Inner Demon, and clicking during the sweep sends nothing.
- **Not tested in game:** the seven other spells listed above.
- Found while testing, not addressed here: the server's `dbc/Spell.dbc` differs from the one in client patch rev4 (`patch-T.MPQ`) on 10 records numerically and 559 enUS texts, and the rev4 file cannot be copied to the server as is (non-enUS string offsets past the string block make `DBCFileLoader::getString` assert).

## Data/source dependencies

No SQL in this change; applied SQL is unchanged. The client half needs `Spell.dbc` regenerated with `python apps/coa-spells/inner_demon_gcd.py --input <client Spell.dbc> --output <new file>` and shipped in the client patch; the generator only changes record 804216 (3 bytes) and never opens an MPQ. No credentials, profiles or game archives are included.

Fixes #391


src/server/game/Spells/Spell.cpp
```diff
@@ -3669,7 +3669,15 @@ SpellCastResult Spell::prepare(SpellCastTargets const* targets, AuraEffect const
     if ((HasTriggeredCastFlag(TRIGGERED_CAST_DIRECTLY) && (!m_spellInfo->IsChanneled() || !m_spellInfo->GetMaxDuration())) ||
         (m_caster->IsPlayer() && m_caster->getClass() == CLASS_NECROMANCER && m_spellInfo->Id == 500991) ||
         (m_caster->IsPlayer() && m_caster->getClass() == CLASS_STARCALLER && m_spellInfo->Id == 800386))
+    {
+        // SPELL_ATTR4_ALLOW_CAST_WHILE_CASTING adds TRIGGERED_CAST_DIRECTLY to ordinary player
+        // casts, which sends them down this branch and past the global cooldown their own
+        // StartRecoveryTime asks for. Real triggered casts carry TRIGGERED_IGNORE_GCD through
+        // TRIGGERED_FULL_MASK and still skip it. Trigger before cast(), which may finish the spell.
+        if (m_spellInfo->HasAttribute(SPELL_ATTR4_ALLOW_CAST_WHILE_CASTING) && !HasTriggeredCastFlag(TRIGGERED_IGNORE_GCD))
```
jealous-sound, line 3677: [P2] Preserve the new cooldown when an older cast is canceled

Start Fel Fireball, wait for its initial global cooldown to expire, cast Inner Demon while Fireball is still casting, then cancel Fireball before Inner Demon's one-second cooldown expires. The direct branch leaves Fireball as the current generic spell, so `Spell::cancel()` calls `CancelGlobalCooldown()` for Fireball; `GlobalCooldownMgr::CancelGlobalCooldown()` clears category 133 without checking which cast started the active cooldown. This removes Inner Demon's newly started cooldown and lets the server accept another ability immediately. Track the originating cast or a generation token for each active global cooldown, and only cancel it when that cast owns it. This sequence is established by source tracing, not a gameplay reproduction.

lostmind84: Reproduced and fixed in eef6f2f. Your sequence let Vengeful Pact through 3/3 on 0eeb57e; it is refused (NOT_READY) 3/3 with the fix.

Each started global cooldown now carries a generation, and a cast only cancels the category if its own generation still owns it. Casts that started no cooldown cancel nothing; callers without a generation (the DK ghoul explosion) keep the unconditional cancel. Regression test: ConquestOfAzerothGhost e2e/classes/felsworn/felfury TestFelsworn_InnerDemonCooldownSurvivesCancel (mattparkes/ConquestOfAzerothGhost#5).

jealous-sound (approved):

PR #1417, opened 2026-09-16
Coa/reaper forms rites and souls

Mrpirolito:
## Data/source dependencies

Confirm that applied SQL is unchanged and that no credentials, profiles or game archives are included.


modules/mod-ascension-compat/src/AscensionCompat.cpp
```diff
@@ -2163,6 +2169,13 @@ class AscensionResourceService
         if (player->getClass() != CLASS_REAPER)
             return;
 
+        // Harvest Time suspends the cost outright rather than rolling for it. Checked before any
+        // list so it covers every path below: the flat consumers, the Soul Infusion requirement and
+        // the single-soul range. An eight second window a Reaper can plan a rotation around is what
+        // the ability is for; a coin flip per cast is not something the player can act on.
+        if (player->HasAura(SPELL_REAPER_HARVEST_TIME))
```
jealous-sound: [P2] Restrict Harvest Time's cost exemption to Soul Infusion abilities

This unconditional return also makes Sanguine Orb (500361) and Tormented Souls (500483) free while Harvest Time is active. Their installed Spell.dbc records require Reaped Soul (CasterAuraSpell = 500363), whereas Harvest Time (803995) explicitly applies to spells that require Soul Infusion (803031). A Reaper with only one soul therefore receives an unintended cost exemption even without Soul Infusion. Please scope preservation to Soul Infusion consumers, leaving ordinary Reaped Soul costs intact even if the deliberate 100% preservation tuning is retained.

Mrpirolito: done, new commits done, can you check again pls?

jealous-sound: Seems good, fix conflict issues and I'll merge

Mrpirolito: done, conflicts fixed

jealous-sound (approved):

PR #1484, opened 2026-09-16
feat(Core/DBC): build the server DBC set from the client players run

jealous-sound:
## Summary

The worldserver's `data/dbc` was a stock 3.3.5a set with about 40 tables copied from the CoA client. Against the client players run (launcher archives plus CoA-Client-Patch), it was missing:

| Table | Missing rows |
|---|---|
| Map / MapDifficulty | 239 / 503 |
| Achievement | 20,787 |
| CreatureDisplayInfo / CreatureModelData | 59,106 / 33,561 |
| GameObjectDisplayInfo | 117,115 |
| ItemExtendedCost / ItemLimitCategory | 12,805 / 2,529 |

The startup log shows the effect: 60,751 items lose their LimitCategory, and all 2,266 of those values exist in the client. 1,225 gameobject spawns are skipped for invalid displayIds, and all 847 of those IDs exist in the client.

### Changes
- **`apps/coa-dbc/client_dbc.py`**
  - `extract`: pulls the client's effective DBC set from `Data/` and `Data/<locale>/` in client load order and writes a manifest of which archive each table came from.
  - `check`: verifies the set against this tree's `DBCfmt.h` and `LOAD_DBC` list.
  - `diff`: compares sets by the values the core reads.
  - `install`: installs a checked set into a DataDir, backing up the previous files.
- **Removes `mod-worldforged-pickups/tools/install_gameobject_display_info.py`.** It merged a single table; the full set already contains it.
- **DBC loader:** CoA's Spell.dbc points 7,029 non-enUS Name/NameSubtext strings (locale slots 9–15) outside its string block. The client never reads those slots, but `DBCFileLoader::Record::getString` asserted on them. `AutoProduceStrings` now loads them as `""`, and `LoadDBC` logs one warning per table.

## Verification
- `python -B apps/coa-dbc/test_client_dbc.py`: 8 tests pass, and the step is added to the CoA quality workflow.
- `modules/mod-ascension-compat/tests/client_compat/run.py` (MSVC): 30/30. It compiles the real `DBCStore.cpp` and `DBCFileLoader.cpp`.
- New gtest `DBCFileLoaderTest.StringOutsideTheBlockLoadsAsEmpty`: compiled standalone with MSVC against googletest and passes. The full `unit_tests` target was not built.
- The real `DBCFileLoader` with `SpellEntryfmt` loads the players' Spell.dbc: 209,365 rows, 7,029 empty strings, spell 17 "Power Word: Shield".
[...]

PR #1487, opened 2026-09-16
fix(DB/DBC): stop SQL overlays from replacing CoA client DBC rows

jealous-sound:
## Summary

A `*_dbc` world table row replaces the loaded client DBC row with the same index. I measured every non-empty overlay against the DBC set the players' client runs (extracted with the tool from #1484). This PR removes the rows that replace or repeat client data.

| Table | Removed | Why |
|---|---|---|
| 11 `gt*_dbc` tables | all 10,274 | Stock ACDB values. They replaced 535 of 3,200 combat rating conversions (for example, at level 51 resilience is 23.8 on the server vs 85 in the client, and rating 25 is 0 vs 0.56), 95 of 100 NPC mana cost scalers, and the crit and regeneration rows of classes 1–11 (for example, spirit mana regen 0.00727 vs 0.00969). Classes 12–32 already read the client. |
| `factiontemplate_dbc` | all 841 | Every ID exists in the client; 6 have different flags, friends or enemies. |
| `spelldifficulty_dbc` | 50 of 604 | These rows remap spells the client already maps to its own difficulty variants. Their IDs are higher, so they are loaded later in `SetSpellDifficultyId` and win. |
| `spell_dbc` | 3,728 of 4,492 | They load the same values as the client row. The only differences are non-enUS name slots and `EffectChainAmplitude` on effects with `ChainTarget <= 1`, which the core only applies to chaining effects (`SelectImplicitChainTargets`). |
| `creaturedisplayinfo_dbc`, `creaturemodeldata_dbc`, `currencytypes_dbc` | 3, 3, 1 | Copies of client rows, added while the server's extract lacked them. |

`DBCStores.cpp` now also stops startup unless DataDir holds the CoA client set. It checks for Rune of Ascension currency 375250, creature display 236827, gameobject display 87226, item limit category 2414 and map 3690. Without the SQL backfills, a stock or partial set would otherwise drop that content silently.

### Still present
- **Additive rows** (IDs the client doesn't have), which stay: 554 `spelldifficulty_dbc` rows, 3 `achievement_dbc` rows, 2 `worldmaparea_dbc` rows and 4 `spell_dbc` rows (46918, 61417, 61418, 74804).
- **760 `spell_dbc` rows** that differ from client spells with the same ID. Core code uses some of them, e.g. 100102 for arena gear scaling in `Map.cpp` and 100101 for Twin Val'kyr, while the client defines those IDs as "Holy Light". Our own 93461 (Adventurer's Cache trigger) also replaces the client's "Dummy Spell". These need the serverside spells moved to free IDs and are a separate PR.

## Verification
- The analysis compares SQL rows with the extracted client set using the value types the core loads.
- A read-only check turned every `DELETE` into `SELECT COUNT(*)` against the local world DB. Each one matches the counts above.
- `codestyle-sql.py` and `codestyle-cpp.py` pass.
- **Not tested:** the C++ change was not compiled, and worldserver startup and gameplay were not run.

## Deployment
**Requires #1484.** Install the client DBC set on every server before starting with this change, or startup exits with `DataDir does not hold the CoA client DBC set`. Stats, regeneration and NPC mana costs for classes 1–11, all combat rating conversions, faction templates and 50 difficulty mappings switch to the client's values.


jealous-sound:
Combined into #1498.

PR #1493, opened 2026-09-16
fix(DB/Spells): use CoA client spells over colliding serverside rows

jealous-sound:
## Summary

After #1487, 760 serverside `spell_dbc` rows still replace spells that the players' client Spell.dbc defines under the same ID. The goal is to recreate the original game's behaviour, so the server now uses the client's spells wherever the client row can actually run the mechanic.

| Change | Rows | Why |
|---|---|---|
| Removed | 643 | The client spell has a full definition. Most are the same helper spells (credit markers, boss dummies, quest triggers) with later AzerothCore edits to targets, radii or attributes. A few are unused AzerothCore helpers covering different client spells, e.g. 11202 Crippling Poison II, 29372 Eruption and 100001 Cauterizing Fire. Our own 93461 row also goes; the client defines it as a dummy spell. |
| Removed | 100099, 100100 | Unused "Stairway Event Spell" rows covering the client's Draconic Warhorn and Hammer of Justice. |
| Moved to 4000000–4000002 | 3 | AzerothCore mechanics that used IDs the client assigns to different spells: arena gear scaling (100102, used in `Map.cpp`) and the Twin Val'kyr ball dummy (100101, plus its `spell_script_names` row) are both "Holy Light" in the client, and Zul'Aman Fixate (43360) is "Call of the Beast". Nothing in the client uses IDs 4000000–4099999. |
| Kept | 111 | The client row has no effects, or fewer effects than the server row, so it cannot run the mechanic. Examples: SetHealth 43536–43538, Halion Meteor Strike 74638/74696, Overlord's Brand 69173, pet scaling 67557/67561, Mingo's Fortune and the Violet Signet creation spells. |
| Kept | 3 | Scripts hook an aura type only the server row has: Garr and Majordomo Separation Anxiety 23487/21094 (periodic dummy, client: dummy) and Infernal scaling 36186 (aura 13, client: 345). |

All 114 kept IDs: 21094 22352 22353 22354 22520 22843 23487 36186 39651–39654 39797 39929 40804–40809 40908 40910–40916 40918–40923 41529–41532 42825 43536–43538 44918–44932 44962 45155–45170 45196 45396 47090 47416 47680 50158–50160 50996 51117–51119 55827–55831 60989 60992 62474 62915 63887–63890 67288 67557 67561 68496 68575 69173 74638 74680–74683 74687–74689 74696

## Method
- Rows were compared with the client set extracted by #1484, using the value types the core loads and ignoring the unused non-enUS locale slots.
- A row counts as the client's "full definition" unless the server row has effects the client row lacks.
- References were scanned in world DB tables (spell_script_names, smart_scripts, spell_linked_spell, conditions, creature spells and auras, items, quests, spell_proc, scripts) and by grepping C++.
- 26 of the colliding spells are bound to scripts. After this change, three scripts target spells whose client rows differ, but their hooks should still bind:
  - `spell_garr_frenzy` (19515): the client adds effect 2, and the hook is on effect 0.
  - `spell_gen_translocate_up`/`_down` (45367/45370): implicit target only; the hook is `AfterCast`.

## Verification
- `codestyle-sql.py` and `codestyle-cpp.py` pass.
- A read-only count shows the `DELETE` list matches 646 existing rows and IDs 4000000–4000002 are unused.
- The relocated `spell_dbc` rows are identical to the ACDB base rows apart from the ID.
- **Not tested:** compiling, worldserver startup and gameplay (arena, Trial of the Crusader, Zul'Aman and the affected bosses).

## Deployment
Requires #1484, whose client DBC set provides the spells these rows used to replace. Independent of #1487.


jealous-sound:
Combined into #1498.

PR #1495, opened 2026-09-16
feat(Core/DBC): add a generic reader for client DBCs the core has no store for

jealous-sound:
## Summary

The core only has stores for the 110 standard DBCs. Server code that needed CoA's custom client tables either had its own reader or relied on data generated offline. This PR adds a generic reader and moves the first consumer, collections, onto it.

**`ClientDBC`** (`src/server/shared/DataStores`)
- Loads a WDBC file once, checks the header against the file size, and requires a minimum number of DWORDs per record.
- Reads records by DWORD index (`GetUInt32`/`GetInt32`/`GetFloat`/`GetString`) or byte offset (`GetUInt8`). Custom tables don't follow the core's format strings: VanityCollection stores 77 DWORDs under a 76-field header, and CharacterAdvancement mixes byte fields between DWORDs.
- Strings pointing outside the string block read as empty.

**`GetClientDBCPath(fileName)`** (`DBCStores.h`) resolves a table in the same directory `LoadDBCStores` read. Custom tables therefore come from the same client set as the core stores (#1484).

**mod-ascension-compat collections** now load Appearances, ItemAppearances, ItemSet and VanityCollection through `ClientDBC` from `DataDir/dbc`.
- Removed: the module's `ForEachWdbcRecord` reader, the `AscensionCompat.DbcDirectory` setting, and the need for a separate `dbc/Ascension/` copy.
- ItemSet is still read raw, because CoA sets list up to 17 items and the core's ItemSet store keeps 10.

## Verification
- New gtest `ClientDBCTest` covers a mixed DWORD/byte layout, strings, an out-of-range string, short records, a truncated file and a missing file. It was compiled standalone with MSVC (`/W4`, no warnings from the new code) against googletest and passes. The full `unit_tests` target was not built.
- The module's four loops, run through `ClientDBC` against the players' client DBC set, match an independent Python read exactly: 42,903 appearances, 202,932 item appearance mappings, 2,280 item sets and 10,764 vanity entries, with a matching checksum over the vanity fields the module uses.
- `codestyle-cpp.py` passes.
- **Not tested:** `AscensionCompat.cpp` was not compiled, and worldserver startup, transmog and vanity were not run in game.

## Deployment
Requires the client DBC set installed by #1484's tool, which puts Appearances.dbc, ItemAppearances.dbc and VanityCollection.dbc in `DataDir/dbc`. Without them, collection data stays unavailable and the error log names the missing file. Remove `AscensionCompat.DbcDirectory` from local configs; it is no longer read.

## Next consumers
- SpellCharges/SpellChargesCategory, replacing the generated 172-entry charge table and hand-coded pools.
- CharacterAdvancement, replacing the 3,618 generated talent entries.


jealous-sound:
Combined into #1498.

PR #3150, opened 2026-09-17
fix(CoA): stop Reaper resource generation depending on the target's m…

Mrpirolito:
…itigation

Reap and Wraithblade generated no Soul Fragments and no Runic Power, and Wraithblade handed out none of its three Reaped Souls, while their ResourceGainRules rows were present and correct the whole time.

An instrumented OnSpellHitResult showed why:

  spell 573303 miss=0 successful=true hostile=true damage=0 crit=true

The damage figure this hook receives is what survived the target's mitigation, and npc_training_dummy::DamageTaken sets damage = 0 on every hit. Every rule on FirstSuccessfulDamagingHit therefore failed its damage test against a training dummy - which is exactly where a player checks a rotation, and where these were being tested.

Resource generation is a property of the ability, not of what the target did with the damage, so read it off the spell: SpellDealsDamage asks the SpellInfo whether it carries a damaging effect, and the four damaging-hit events now qualify on that or on a non-zero figure. A spell with no damaging effect still does not qualify, so nothing that was not generating starts to.

Verified in game against a Grandmaster's Training Dummy: fragments climb 0, 1, 2 and convert to a Reaped Soul at three, Soul Infusion appears at three souls, and Wraithblade grants its three souls and the infusion with them.

## Data/source dependencies

Confirm that applied SQL is unchanged and that no credentials, profiles or game archives are included.


jealous-sound:
Reviewed commit dfc4861425e92b035b1cb27a4fa5e9510cda0c9b.

**[P2] The new helper breaks the resource-generation test harness.**

[`AscensionCompat.cpp:1801`](https://github.com/Mrpirolito/azerothcore-wotlk-coa/blob/dfc4861425e92b035b1cb27a4fa5e9510cda0c9b/modules/mod-ascension-compat/src/AscensionCompat.cpp#L1801) calls `SpellDealsDamage`, but `modules/mod-ascension-compat/tests/resource_generation/run.py` extracts `OnSpellHitResult` without extracting the new helper. Its `SpellInfo` fixture in `harness.cpp` also lacks `HasEffect` and the spell-effect constants needed by the helper. The generated C++ therefore cannot compile. Update the extraction and fixture, and add coverage for zero-damage damaging hits, misses, and non-damaging spells.

Verification: inspected the generated method selection and confirmed the missing helper and fixture API. Scoped C++ lint, `git diff --check`, and GitHub CI passed. The training-dummy diagnosis matches the source. No build or in-game verification was performed.

Mrpirolito:
done

Tareksoh:
**Live-log check of this gate (108 retained live combat logs, 3.3.5 format)**

I checked the retained live logs against the two behaviours this gate decides, for every Reap and Wraithblade cast by a player (Reap 2,582 casts: ranks 500357, 504056-504058, 573302, 573303; Wraithblade 158 casts: 805258, 806822, 806823). For each cast I took that cast's SPELL_DAMAGE or SPELL_MISSED and looked for Soul Fragment 805077, Reaped Soul 500363 or Soul Infusion 803031 applied to the caster within 0.5 s, third-stack conversions included. "Capped" below means the caster already held 3 Reaped Souls with Soul Infusion up; live generated almost nothing in that state (3 of 483 hits, 0 of 19 misses).

**Absorbed to zero: no live data.** None of the 2,495 Reap hits or 169 Wraithblade hits has a non-zero absorbed field or an amount of 0, and there is no `SPELL_MISSED ... ABSORB` for either spell. The same files hold 25,703 ABSORB misses and 23,866 partially absorbed damage rows for other spells, and no Reap in them struck a player target, so most likely no shield was met. Wraithblade in addition ignores absorption by its own tooltip, so an absorbed-to-zero hit cannot occur for it at all. The new zero-damage path is neither confirmed nor contradicted by live; I mention it only so nobody assumes a measurement exists for it.

**Misses: live did not gate Reap's fragment on the hit landing.**

| Reap result | casts | fragment gained | not capped | capped |
|---|---:|---:|---:|---:|
| hit (amount > 0) | 2,495 | 1,995 | 1,992 / 2,012 | 3 / 483 |
| dodge | 45 | 34 | 34 / 34 | 0 / 11 |
| parry | 28 | 25 | 25 / 25 | 0 / 3 |
| miss | 13 | 8 | 8 / 8 | 0 / 5 |
| immune | 1 | 1 | 1 / 1 | - |

All 68 uncapped dodge, parry, miss and immune fragments arrive within 0.02 s of the SPELL_CAST_SUCCESS, like the hits (1,760 of the 1,985 direct fragment gains at +0.00 s), from 15 different casters. 23 of them have no other cast, hit or swing by the caster within 0.5 s, and a control window 0.05-0.45 s before the same 87 casts holds 3 fragments, so coincidence is out. On live the fragment text belongs to the Soul Collector passive 706731 ("Your Reap generates 1 Soul Fragment"), so this is that passive firing on the Reap attempt rather than on the landed hit; the log cannot show which spell applied the aura. Of the 20 uncapped hits without a fragment, 9 have no visible prior state in the log and 2 are conversions this count missed. Runic Power cannot be checked this way: there are 0 energize events in 2,582 casts, none for 355461 either, although the tooltip says Reap generates Runic Power.

This PR does not change the `successful` gate, so nothing here blocks it. It does mean the "a miss is still a miss" case in `tests/resource_generation/cases.cpp` encodes something live did not do for Reap and Soul Fragment 805077. Whether to keep that gate or follow live is your call. For reference, at 3 Reaped Souls with Soul Infusion up, live granted no fragment on any of 19 misses and on 3 of 483 hits.

**Wraithblade:** all 158 casts landed (0 misses; absorbs cannot occur for it by design, see above), so there is no live data on the miss question for it. 147 of 148 uncapped casts gained Reaped Soul and Soul Infusion in the same instant; 8 casts hit 2-3 targets and granted once each, consistent with a `FirstSuccessfulDamagingHit` event.

Method: two-pass parse; a result is attributed to a cast when it is the same spell family by the same caster within 0.6 s and before the caster's next cast of it; player identities were discarded before analysis.

jealous-sound (approved):

PR #3671, opened 2026-09-17
fix(Tinker): prevent sentry from autonomously pulling hostiles

kiwispin:

Tinker Sentry Turrets could autonomously acquire nearby hostile units that the player had not attacked.

The fallback targeting logic scanned nearby units and accepted targets based on combat/hostile state. This could cause a Sentry to pull additional mobs on its own.

This change removes autonomous nearby target acquisition. The Sentry now assists targets explicitly attacked by its owner, switches when the owner deliberately attacks a new target, and remains idle when no valid player-initiated target exists.

- Added regression coverage for idle nearby hostiles, incoming aggro, explicit player attacks, target switching, and selected-but-not-attacked targets.
- The C++ Tinker Sentry regression harness compiles and executes successfully on Windows.
- The existing Python runner exits afterward because Windows temporarily keeps `sentry.exe` locked during cleanup, so the runner does not reach its final PASS message.
- The behavior has also been tested in-game on the local CoA server.

AI-assisted: ChatGPT and Codex were used to investigate, implement, and test this change.

## Data/source dependencies

No SQL, DBC, client, credentials, profiles, or game archives are changed or included.


modules/mod-ascension-compat/src/AscensionTinkerSummons.cpp
```diff
@@ -370,19 +370,12 @@ struct npc_ascension_tinker_device : ScriptedAI
...
+        // Explicit player-initiated Tinker attacks propagate their focus through SetGUID.
+        if (Unit* target = ObjectAccessor::GetUnit(*me,focus); valid(target))
             return target;
-        Unit* nearest = nullptr;
-        for (Unit* target : Nearby(me,range))
-            if (valid(target) && (player->IsInCombatWith(target) || player->IsHostileTo(target)) &&
-                (!nearest || me->GetExactDist(target) < me->GetExactDist(nearest)))
-                nearest = target;
-        return nearest;
+        return nullptr;
```
jealous-sound: [P2] Preserve assistance for other spell attacks

The focus update in `tinker_spells::OnSpellCast` only handles Scrap Shot and Gatling Gun. If the player summons an idle turret, then attacks with Rocket Launcher without starting melee, `GetVictim()` and `focus` remain empty, so the turret no longer assists. Removing the selection fallback needs a replacement that records explicit spell/ranged attacks. Add coverage through that attack-notification path.

modules/mod-ascension-compat/src/AscensionTinkerSummons.cpp
```diff
@@ -370,19 +370,12 @@ struct npc_ascension_tinker_device : ScriptedAI
...
         };
-        if (Unit* target = ObjectAccessor::GetUnit(*me,focus); valid(target))
-            return target;
         if (Unit* target = player->GetVictim(); valid(target))
             return target;
-        // Spell and ranged attacks need not set the player's melee victim.
-        if (Unit* target = player->GetSelectedUnit(); valid(target) && player->IsInCombatWith(target))
+        // Explicit player-initiated Tinker attacks propagate their focus through SetGUID.
+        if (Unit* target = ObjectAccessor::GetUnit(*me,focus); valid(target))
             return target;
```
jealous-sound: [P2] Do not let an older melee victim override a new spell target

Unconditionally preferring `GetVictim()` breaks switching in the opposite direction from the new test: melee-attack A, then cast Scrap Shot on B through an explicit-target macro. The cast updates `focus` to B without changing the melee victim, so the turret keeps shooting A. Track which player attack is newer instead of always prioritizing the melee victim, and cover this switching direction.

modules/mod-ascension-compat/src/AscensionTinkerSummons.cpp
kiwispin: Addressed in adfae2f. Explicit non-triggered hostile Tinker spell casts now update the shared attack focus generically rather than only handling Scrap Shot/Gatling Gun. Added harness coverage for the Rocket Launcher-style case as well.

modules/mod-ascension-compat/src/AscensionTinkerSummons.cpp
kiwispin: Addressed in adfae2f. The turret now follows a shared most-recent attack focus: explicit spell attacks update it immediately, while changes to the player's actual attack victim update it when a newer melee/ranged attack occurs. Added coverage for melee A -> spell B and spell B -> new melee C.

modules/mod-ascension-compat/src/AscensionTinker.cpp
```diff
@@ -40,6 +40,33 @@ TinkerState& State(Player* player)
...
+{
+    if (!player || !target || !player->IsValidAttackTarget(target))
+        return false;
+    State(player).focus = target->GetGUID();
+    return true;
+}
+bool NotifySpellAttack(Player* player, SpellInfo const* spellInfo, Unit* target)
+{
+    if (!spellInfo || spellInfo->SpellFamilyName != 34 || spellInfo->IsPositive() || !NotifyAttack(player,target))
+        return false;
```
jealous-sound, line 53: [P2] Auto Shot still does not update turret focus

Auto Shot (75), available to Tinkers, belongs to spell family 9, so this family-34 filter rejects it. Its native auto-repeat path also leaves `GetVictim()` unset, meaning `ObserveAttack()` cannot detect it. Opening with Auto Shot therefore leaves the turret idle; switching Auto Shot targets leaves its previous focus unchanged. Handle the native ranged-attack notification and test with an auto-repeat target while `GetVictim()` is null—the current harness simulates ranged attacks by assigning a melee victim.

Can you please verify it?

kiwispin: Addressed in 28c2a12. Native ranged auto-repeat targeting is now observed via CURRENT_AUTOREPEAT_SPELL and its unit target, independently of GetVictim(). Added regression coverage with GetVictim() == nullptr for Auto Shot A, A → B switching, and stale Auto Shot not overriding a newer melee target.

jealous-sound (approved):

PR #3993, opened 2026-09-17
CoA: plate-class weapons, Invincible, level scaling, Reaper Runic Power, guide trainers, chat links and the Reaper tier set

Mrpirolito:
Everything this fork's realm has been running locally that is finished, minus the Reaper tier set, which stays out until it is polished.

Five independent commits; each one stands alone and can be dropped without touching the others.

---

## 1. `fix(Core)` — the fork's plate classes and the Strength weapons and trinkets

**Files:** `src/server/scripts/Spells/spell_item.cpp`, `src/server/shared/SharedDefines.h`, `data/sql/updates/pending_db_world/rev_20260916_64_shadowmourne_plate_classes.sql`

Deathbringer's Will and Aura of Madness choose their proc from a table indexed by class, and those tables end at `CLASS_DRUID`. Every Ascension class (12–32) indexed past the end, so both trinkets did nothing at all for them.

`LegacyTrinketTableRowFor(Unit const*)` maps a custom class onto the legacy row it plays like. Plate wearers map onto `CLASS_WARRIOR`, which is the Strength row; everything else goes through the existing `GetLegacyClassForCustomClass`.

`AscensionClassWearsPlate(Classes)` is a `constexpr` next to `IsAscensionClass` rather than a local helper, because "what armour does this class wear" is a property of the class list and the Shadowmourne change below needs the same answer.

Shadowmourne was gated to Warrior, Paladin and Death Knight — the three plate classes WotLK shipped with — so a Knight of Xoroth could not even start *The Sacred and the Corrupt*. The gate should be "can this class wield the reward", which is the Two-Handed Axes proficiency (spell 197) that `ClassProficiencies` grants, intersected with actually wearing the armour:

* **In:** Knight of Xoroth (17), Cultist (25), Reaper (30), and Barbarian (12) — its Body Builder passive converts Strength to attack power, so it has a real Strength build to spend the axe on.
* **Out, deliberately:** Starcaller (26), asked for by the realm despite holding the proficiency; Runemaster (32), a leather class; Guardian (18), Sun Cleric (27) and Primalist (31), which wear plate but have no two-handed axe proficiency, so the reward would not be equippable.

The migration moves the whole chain, the axe, Frostmourne and Shadow's Edge together, so the chain cannot complete into an item the character cannot equip. Shadow's Edge shipped as `-1501`, which silently included all twenty-one custom classes.

## 2. `fix(DB)` — Invincible's two missing effect types

**File:** `data/sql/updates/pending_db_world/rev_20260916_67_invincible_mount_effects.sql`

Invincible (72286) could not be cast: the client answered *"You have to be unsheathed to do that!"* and the server never saw a cast packet. Two fields on the spell were damaged.

* `Effect_1` and `Effect_2` were `0` instead of `6` (`SPELL_EFFECT_APPLY_AURA`), so the spell applied nothing.
* `AuraInterruptFlags` was `0x0047FF7E`, which carries `AURA_INTERRUPT_FLAG_NOT_SHEATHED` (`0x200`). No other mount wrapper of the same shape carries it. The value now matches 39798, a working mount built the same way.

Verified in game: the mount summons and the character rides it.

> The same two fields have to be right in the client's `Spell.dbc` for the tooltip and the cast to agree. That is a client patch, not part of this PR.

## 3. `feat(CoA)` — an in-world switch for creature and quest level scaling

**Files:** `modules/mod-ascension-compat/src/AscensionLevelScalingToggle.cpp`, `modules/mod-ascension-compat/src/MP_loader.cpp`, `data/sql/updates/pending_db_world/rev_20260916_63_level_scaling_toggle_npc.sql`, `..._69_orgrimmar_entrance_placements.sql`

`LocalLevelScaling::CreatureEnabled` and `QuestEnabled` are process-wide switches read once from `mod_ascension_compat.conf`, so answering "is this fight hard because of scaling, or because of me?" meant editing a file and restarting the world.

The **Keeper of Proportion** (990010) flips both from a gossip menu and remembers the answer in `worldstates` 2100001 / 2100002. `0` means "never set", so the configured value still wins on a fresh realm and "off" stays distinguishable from "unset". `OnStartup` re-applies the stored choice, because `OnBeforeConfigLoad` writes the configured values before world states are loaded.

One switch, not two: creature and quest scaling are halves of the same thing — a world that rises to meet the player — and a realm wanting one without the other has not come up. Creatures already spawned keep the level they spawned with either way; the switch applies from the next spawn, and the gossip says so by reopening with the new state.

The second migration places it at the Valley of Strength entrance beside a pair of training dummies (31144 and 31146). Both ground heights come from spawns that already exist there, not from a guess.

## 4. `fix(CoA)` — the Reaper's Runic Power builders

**File:** `modules/mod-ascension-compat/src/AscensionCustomResourceData.h`

Soulrend carried a `150` placeholder and Doomrend had no row at all, so casting it built nothing. Both now grant 20 Runic Power — `200` in the tenths this table uses.

The comment states plainly what the number is and is not: the tooltips say only "generates Runic Power", the retained live logs carry **no** energize event for either because the gain was applied silently, and spend-over-cap windows only bound it from below. The figure comes from the realm this was written for, not from a log, and the comment says so rather than claiming parity.

573320 (Soulrend's armor debuff) and 560361 / 800173 (Doomrend's healing-absorb and heal halves) stay outside the ranges so they cannot grant a second time.

## 5. `feat(CoA)` — guide trainers, Darion's runeblade exchange, and login state that survives

**Files:** `modules/mod-ascension-compat/src/AscensionCompat.cpp`, `src/server/game/Globals/ObjectMgr.h`, `src/server/game/Entities/Creature/Creature.h`, `apps/client-compat/CoA_TalentStateFix/`

Five changes grouped because they share the login path.

**`AscensionGuideTrainer`** turns every Book of Ascension companion into a real trainer: the player's own class abilities plus the full profession list, gated by their real skill requirements. It is an `AllCreatureScript` so it claims the gossip before `npc_ascension_training_book`'s `CreatureScript` sees it, and it carries that script's "restore my available abilities" option so nothing is lost. Buying goes through `CanPacketReceive` on `CMSG_TRAINER_BUY_SPELL` — on the **map thread**. `CanPacketReceiveEarly` runs on the network thread and teaching a spell there crashes the world. `ObjectMgr::GetTrainerById` comes with it, because the native lookup resolves trainers by creature entry and a companion is not one.

**`AscensionRunebladeReforge`** is Highlord Darion Mograine's free, repeatable exchange between Shadowmourne (49623) and Frostmourne (33350), gated behind the completed chain. It rewrites the entry of the item instance the player already owns instead of destroying and granting a new one, so gems, enchantments, durability, the soulbound flag and the item GUID all survive.

**`SendOwnedTalentSpells`** reports the character's owned talents and active specialization to the client at login; `.localtalentsync` repeats it for a session that reconnected without one. Without it the client draws an empty tree for a custom class. The addon that consumes the message is in `apps/client-compat/CoA_TalentStateFix/`, with its own README.

**The login resync no longer refunds a tree.** The stored specialization is also set from the first spec-bound talent a character ever picks, which need not be the tab they actually play. When the client then announced its real spec at login, the mismatch was answered with a full refund and the build was gone. Inside the grace window the client's value simply wins, and the adoption is logged.

**`RestoreActionButtonsForGrantedSpells`** re-reads `character_action` at login and puts back the buttons for spells granted after the action bar was loaded. `Player::_LoadActions` runs inside `LoadFromDB`, before those grants, so `IsActionButtonDataValid` rejected them and every such button silently disappeared. It uses a direct query on purpose: `CHAR_SEL_CHARACTER_ACTIONS_SPEC` is prepared for `CONNECTION_ASYNC` only, and running it synchronously trips `ASSERTION FAILED ... MySQLConnection.cpp:261 Condition: m_mStmt` on every login.

**`MAX_VENDOR_ITEMS`** is unchanged at 150 and only gains a comment. The packet's own count is a `uint8`, so 255 looks like the ceiling; it is not. Raising it and handing a vendor 160 items crashed the client the moment the list opened. A vendor with more to sell gets a second vendor beside it.

## 6. `feat(CoA)` - the AtlasLoot chat link fix, as an addon

**Files:** `apps/client-compat/CoA_ChatLinkFix/`

A shift-clicked item link cannot be used in a dot command on this client, and both reasons are in
the link the client builds rather than in the command. It ends in `|r|r` - the second colour
terminator closes nothing, the client refuses the whole message with *"Invalid escape code in chat
message"*, and the `.additem` never leaves the edit box. And the link itself is the short form,
`|Hitem:50363|h[Name]|h|r` with none of the trailing item fields, which the server's hyperlink
parser answers with *"Provided link has invalid link data"*.

Commands take a bare item id just as happily. The addon collapses runs of terminators in every
message and, in a message starting with a dot only, reduces each item link to its id. Ordinary chat
keeps its links.

## 7. `feat(CoA)` - the Reaper's Icecrown tier set

**Files:** `rev_20260916_65`, `_66`, `rev_20260917_00`, `rev_20260920_10`, `_12`, and the set bonus
half of `AscensionCompat.cpp`

Frostbitten Battleplate of the Risen Nightmare: five plate pieces at four item levels - 251, 264
Sanctified, 277 Mythic and 284 Ascended - sold by a quartermaster of its own. The Reaper is a
Strength plate class with no tier of its own, so every stat is copied verbatim from the Ymirjar
Lord's piece of the same slot and item level rather than scaled by a curve invented for the
occasion.

Two set bonuses, both of which need the module because no DBC record can express them:

* **Two pieces** take thirty seconds off Harvest Time.
* **Four pieces** grant Harvested Might, +5% Strength for eight seconds and up to three stacks,
  every time a cast actually spends a Reaped Soul. `ConsumeReaperSouls` now reports whether
  anything was spent, so the caller knows when the bonus fires.

The set gets its own vendor rather than joining Horace Hunderland, who already carries 140 items:
twenty more put him at 160 and opening him crashed the client outright. That is the same lesson
`MAX_VENDOR_ITEMS` keeps its comment about.

**PvE Power and the green source line both live in `item_template`**, which took a while to find.
Comparing 51227, Sanctified Ymirjar Lord's Helmet, against ours shows exactly what the tier carries
and we did not: `description = '@Ascended Raid@'`, and `spellid_4 = 101638` with
`spelltrigger_4 = 1`, which is "PvE Power (+38)" on equip. PvE Power goes by **slot**, not by item
level - head, chest and legs carry 101638 and shoulders and hands 101625 across all three Ymirjar
tiers - so the same two spells cover all four of ours. The source line only draws for an item
carrying flag `8`; all 174,055 items with an `@...@` description have it.

---

## Deliberately not here

* **Talent and ability fixes.** Extinction, Ruin and Redshade are the same class of defect and are in their own pull request.
* **Client-side DBC rows.** The tier set needs four `ItemSet.dbc` rows and three spell records (2990001-2990003, the two bonuses and Harvested Might) in the client's own tables, and Invincible needs its two effect types there as well. Those are a client patch, not a server change, and are not in this branch.

## Testing

Everything above has been running on a local realm and was checked in game: the trinkets proc for a Reaper, the Shadowmourne chain opens and the exchange returns an item with its enchants intact, Invincible mounts, the Keeper of Proportion flips scaling and the choice survives a restart, Soulrend and Doomrend build Runic Power, guide companions train, and a character that logs out with a granted spell on the bar logs back in with it still there.


Mrpirolito:
Everything tested in local and working correctly

Issue #3971, opened 2026-09-17
CoA talents: Reset Trees leaves hidden paid passives active despite empty talent nodes (all classes)


Tareksoh:
### Core revision and client profile

- PTR, core revision `d212cbdb0bb8` (revision recorded for the session; no fresh revision screenshot).
- Inspected client asset: `Interface/AddOns/Ascension_Collections/CharacterAdvancementStateCompat.lua` in `patch-B.MPQ`, 22,161 bytes; SHA-256 `cbfd86eb99e868f25709e4f43bdc460cafefa750c29a9e352c88dc6ca87c4612`. The extracted copy was checked against the archive. No hash was captured during the PTR session itself.
- Reaper (class 30), level 60; talents selected by hand earlier in the session. This was not a clean single-pass reproduction from a fresh login.

### Reproduction, expected result and actual result

**Recorded PTR sequence:** Soulbender (entry 6755, spell 804004) and Dominion (entry 7271, spell 803999) were already learned.

1. Click Reset Trees, then Save Changes, with nothing selected. `.list auras` still shows 804004 and 803999.
2. Run `/reload`. Both talent nodes appear unlearned.
3. Run `.localtalent 6755 0` and `.localtalent 7271 0`. Both auras disappear.
4. Rebuild the trees without those talents. The two auras do not return in the recorded session; other passives whose nodes were not selected remain listed, including Mawborn 805203.

**Expected:** the window reflects the character's learned ranks, and Reset Trees + Save removes paid talents and their effects on the server.

**Actual:** the window shows those nodes unlearned while their auras remain on the server after the UI reset. Direct rank-zero commands remove them. Empty nodes after a reset are expected and do not independently prove that reload lost the ranks. This was observed on one Reaper; a fresh learn → reload → reset sequence and a controlled full-relog check were not separately recorded.

The white-swing proc behavior is separate: #1135 and #1339 concern these spell IDs, but describe missing handlers rather than the observed over-firing.

### Minimal redacted evidence and verification limits

**Static client evidence, re-checked; not a runtime API or packet trace:**

- Both rank spells have `SPELL_ATTR0_DO_NOT_DISPLAY` (`0x80`) in the inspected shutdown, original-client and patched-client spell data.
[...]

Issue #3672, opened 2026-09-17
Passive Talents Not Saving (Stormbringer Class)


jealous-sound:
Submitted from the in-game bug-report form.

#### Spell
 All Passive Talents [ID: Not Included]

#### Issue
Passive talents on my Stormbringer character are not saving. Spells that I have to use (E.G: Arm of Thorim/CloudBurst) correctly save but passive talents like intellect or pushback resistance are not. I've tried deactivating all my addons and repeating the process several times, but passive talents are not saving when talented.
#### Is this a Gamebreaking Issue
Yes
#### Exact Location
Everywhere/Bloodmyst Isle Currently
#### What were you doing
Questing
#### Expected Outcome
Talents to remain
#### Actual Outcome
Talents do not remain
#### Steps to Reproduce
Lvel up a Stormbringer class past level 10. Talent useable and passive spells. Uee /reload ui or relog. You'll see the passive talents have returned to their untalented states.

### Classification
Category: 1
Severity: 3

Client: 3.3.5 (12340)

### Server context
Class ID: 16
Level: 17
Map ID: 530
Position: -2055.29, -11894.4, 46.1488
Core revision: be725043c415

Issue #1397, opened 2026-09-16
Talent Allocations Are Not Saved After Exiting and Re-entering the Game


xpoo1xp:
### Core revision and client profile

Every time I exit the game and re-enter, the talent page does not save my allocations, so I have to assign the talent points again.

### Reproduction, expected result and actual result

Reproduction:

Log into the game.
Open the talent page.
Allocate talent points.
Exit the game.
Log in again.
Open the talent page.
Expected result: The previously allocated talents should be saved and displayed correctly. The player should not need to re-allocate talent points.

Actual result: The talent allocations are not saved. The talent page resets after re-entering the game, and the player has to allocate talent points again.

Issue #100, opened 2026-09-13
Witch Hunter and Knight of Xoroth Talents Resetting on Login


jealous-sound:
Submitted from the in-game bug-report form.

#### Spell
 Combusting Blade [ID: 704999]

#### Issue
When I log in on my Witch Hunter or Knight of Xoroth, I have to choose my spec again and all of their talents (excluding the first one in class and spec tree) reset.
#### Is this a Gamebreaking Issue
Yes
#### Exact Location
Open World
#### What were you doing
Logging in on my Knight of Xoroth and Witch Hunter
#### Expected Outcome
I don't need to choose my specialization and talents again
#### Actual Outcome
I have to choose my spec and talents again
#### Steps to Reproduce
Step 1: Get to level 10+
Step 2: Choose a specialization and talents
Step 3: Relog

### Classification
Category: 1
Severity: 3

Client: 3.3.5 (12340)

### Server context
Class ID: 17
Level: 13
Map ID: 530
Position: 8719.76, -6624.01, 71.4942
Core revision: unknown

### Consolidated duplicate reports

Additional reports: #115 reproduces this on Pyromancer: passive talents and specialization are forgotten after logout/login while active-skill talents remain. #183 reports that talents revert on most relogs. #189 reports the same after login or disconnect and additionally losing the selected stance; preserve that stance symptom for investigation. This issue remains open for all these persistence reports.

lostmind84:
The client-patch fix for this is tracked in PR #4335 (docs handoff); it documents but doesn't close it. No further server-side change here.

PR #4027, opened 2026-09-18
fix(CoA): make the server the authority on CoA talent state

lostmind84:

CoA talent state lived only in the client. The server learned each `.localtalent <entry> <rank>` as a spell,
enforced no point budget, and sent the client's character-advancement service nothing but a zeroed
`0x0725`. The shipped compat addon rebuilt ranks from `IsSpellKnown`, which cannot see
`SPELL_ATTR0_DO_NOT_DISPLAY` passives (2,394 of the 3,219 paid entries), so after any reload or relog those
ranks read as 0: Reset Trees never sent their rank-0 change, the auras stayed active (#3971), and the freed
points could be spent again — the server accepted every request (#3971 follow-up: 28+ points held by a
26-point character). The same client rebuild is why passives look lost after a relog in #100, #1397 and
#3672, and the specialization overlay returns on every login because the addon reads its per-character
SavedVariable before WoW restores it (`CharacterAdvancementCompat.lua:21`); those two client defects belong to
the client patch and are described below.

This PR makes the server the authority:

- `AscensionCoATalentData` also loads `CharacterAdvancementEssence.dbc`: the cumulative class-tree (AE) and
  specialization-tree (TE) points of every custom class per level (key = class id, flags clear; level 10 → 1/0,
  level 60 → 26/25), the same curve the client computes locally.
- `AscensionCoATalentState` derives held ranks from the spellbook, prices them (a rank spell two entries share
  is charged once) and serializes the service's packets.
- `.localtalent` goes through one `SetTalentRank` with every previous rule plus the budget: a rank above the
  held one is refused when the tree cannot pay for it; rank 0 and lower ranks always go through. New
  `.localtalent reset` removes every paid rank of the class and keeps automatic grants and the specialization.
- The service receives its state: `SMSG 0x0725` (slot 0 of 1) on the first `CMSG_SET_ACTIVE_MOVER` of the
  session — the client keys this state off its local player object, which its loading screen has not created
  when `OnPlayerLogin` runs — followed by `SMSG 0x0726`, the complete set of held entries with ranks, resent
  after every `.localtalent`, `.localtalent reset`, `.localspec` and level change. `IsKnownID` /
  `GetTalentRankByID` answer from that set, so a client can show hidden ranks without seeing their spells.
- `CMSG 0x0727`, the service's complete wanted set after a native learn or unlearn, arrives on the network
  thread and is queued for the player's own update, then applied through the same rules: every entry checked,
  the resulting state priced first (including ranks the progression pass hands back to omitted entries), the
  specialization selected when none is active (one per upload), removals before additions, the whole upload
  refused on any failure, and always answered with the server's state. A missing budget row refuses rather
  than allows.

- Folded in from #4030 (jealous-sound): the `ASC_LOCAL_CAD` bridge — an addon-channel whisper carrying the
  active specialization id and the held `entry,rank` pairs, chunked, sent at login and again with every
  known-entries resend — which carries the one thing the native packets cannot, the `ChrSpecs` id, and which
  #4030's patch-B client half already consumes; and the opcode identity corrections (`0x0523` is
  `CMSG_CUSTOM_ASCENSION_POINT_SPEND_REQUEST`, not a vanity delivery; `0x061A` is `CMSG_CREATURE_QUERY_BULK`;
  `0x064A` named), kept as its own commit with its author.

- Folded in from #4031 (bozo-1): **stored builds** — Ascension keeps a build per specialization and swaps
  between them, so a switch now writes down the build being left (class tree and the specialization's own
  tree, as player settings `core.ascension_build.<spec>`) and puts back the build of the specialization
  being entered, each rank through `SetTalentRank` so nothing unaffordable is granted; the record is read
  only on entry, never at login. Also its three-message state form (`ASC_LOCAL_SPEC` / `ASC_LOCAL_RECORDS` /
  `ASC_LOCAL_TALENTS`, sent with the bridge) and `.localspecstate`, so its patch-B client half works
  unchanged too; `.localtalent sync` is the same request. One of the two message forms should retire with
  the client patch that ships. Not taken from #4031: `0x0725` as `(slot, specId)`, sending state from the
  network thread on `CMSG_EXTENSION_INITIALIZED`, the refuse-once `.localspec` guard (a mistaken switch is
  now reversible), and its unused `RefuseContradictingUnlearn`.

Packet layouts follow the community measurements of the Ascension client in
`hertigservices/Ascension_preservation` (`server/world_server.py`, opcode notes for 0x725/0x726/0x727). How a
specialization slot maps to a `ChrSpecs` id and which CMSG a native specialization switch emits are not
documented there, so `0x0725` carries slot 0 of 1 and the specialization id travels in the bridge message
and `.localspec`.

Documentation: `.agents/docs/systems/ascension-talents.md` (new, routed from `AGENTS.md`).

Client side (coa-local-kit, not this repository; proposed diff posted on #3971) —
read the SavedVariable on `ADDON_LOADED` instead of at file scope, take `IsSpellIDKnown` and
`GetTalentRankByID` into the rank fallback. Until it ships, players can `.localtalent reset` (chat) to clear
stuck passives; the budget holds regardless of the client.

Supersedes #4030 and #4031 (their pieces are in here). Refs #3971, #100, #1397, #3672 — left open on purpose: the visible symptom also needs the client patch.

Linux, Docker slot with the client DBC set (`coa-slot preflight` PASS: worldserver built from HEAD, shipped
SQL applied, 368/368 client tables identical):

- `python3 modules/mod-ascension-compat/tests/talent_state/run.py --dbc-dir /srv/coa/server-data/dbc`: 23
  checks pass (budgets 1/0, 1/1, 26/25, 36/35 for classes 12–32, rank derivation, shared-spell charging,
  packet layout, upload parsing incl. malformed sizes).
- Ghost e2e (lostmind84/ConquestOfAzerothGhost#14, merged, and lostmind84/ConquestOfAzerothGhost#15): `TestTalentStateSurvivesRelog`,
  `TestTalentBudgetIsEnforced` (class and specialization trees at level 12: 2/1 points, third rank refused,
  accepted after freeing a point), `TestTalentResetRemovesPaidRanks`, `TestKnownEntriesAreSent` (0x0725 then
  0x0726 on activation, entry appears at rank 1 and disappears at rank 0), `TestKnownEntriesUploadIsApplied`
  (apply, diff, over-budget refusal with state unchanged), `TestKnownEntriesUploadSelectsOneSpecialization`
  (a two-specialization set refused, a one-specialization set selects it), `TestTalentBridgeMessageIsSent`
  (the `ASC_LOCAL_CAD` whisper after a change names the specialization, the paid and an automatic entry, and
  the three-message form carries the paid ranks only), `TestBuildSurvivesSpecializationSwitch` (builds of
  two specializations follow their switches across a relog, class tree kept) — all pass; the existing
  `TestTalentReplacements` still passes.
- `python3 apps/codestyle/codestyle-cpp.py --files <changed>`: the only findings are pre-existing brace
  placements in creature-preset code (also reported on `origin/main`). `git diff --check` clean.

Not verified here: the real client's reaction to 0x0725/0x0726/0x0727 (no lab client run in this PR — the
compat addon overrides the native API, so today's client behaves as before; the packets matter once the
client patch defers to them). The bridge message and its patch-B client half were verified in game by #4030
with `0x0725` sent as `(specId, 0)` at login; this PR sends `(0, 1)` on the first `CMSG_SET_ACTIVE_MOVER`
instead, which #4030's client half does not depend on, but that combination has not been run on a real
client yet. Examined but not changed: 29 paid rank spells are also class progression
grants and read as held ranks (as the client already displays them); two rank spells are shared by two
entries; an automatic entry depending on a removed paid entry is not removed by a reset (same as
`.localtalent <entry> 0` today).

## Data/source dependencies

Reads `CharacterAdvancementEssence.dbc` from the client DBC set the server already loads (#1498); no SQL
change. Applied SQL is unchanged; no credentials, profiles or game archives are included.


lostmind84:
Updated after the overlap with #4030: its opcode-identity commit is now the first commit here (author kept) and its `ASC_LOCAL_CAD` bridge message is folded into the authority commit (sent at login and after every talent change, same format, so its patch-B client half applies unchanged). Rebased on current `main`; the seven Ghost scenarios (#14 merged, #15 open) and the existing `TestTalentReplacements` pass on the rebuilt slot, preflight PASS. Description updated.

lostmind84:
Updated with the parts of #4031 worth keeping: stored builds per specialization (a switch writes down the build it leaves and restores the one it enters, class tree shared, ranks re-checked through the same rules), its three-message state form (`ASC_LOCAL_SPEC/RECORDS/TALENTS`) and `.localspecstate` so its patch-B client half works unchanged, alongside #4030's `ASC_LOCAL_CAD`. Not taken and why: in the description. Eight Ghost scenarios + `TestTalentReplacements` pass on the rebuilt slot (preflight PASS), including a new one that switches 56→55, relogs, and finds each build waiting on its specialization.

jealous-sound (approved):

PR #4030, opened 2026-09-18
fix(CoA): make the server authoritative for the active specialization and paid talent ranks

jealous-sound:
## What was wrong

Three reported symptoms, two root causes.

**The specialization chooser came back after every logout.** The active specialization lived
only in a per-character SavedVariable in the patch-B compatibility layer, and that layer
cleared it whenever its own validation failed — a check that needs the player object and the
local catalogue, neither of which is ready early in a session.

It was this module that triggered the clear. `SMSG 0x0725` was being sent at every login with
two zero dwords. That opcode carries no name in the client's table, but it is a registered
handler: it allocates the Character Advancement state object and its first dword is the active
specialization. Sending zero told the client on every login that the character had none, which
raised the change event, which ran the validation, which cleared the saved id.

**Talent points read as unspent while their spells stayed learned.** The layer reconstructed
paid ranks by looking each entry's rank spells up in the spellbook. Any rank whose spell it
could not match came back as unspent.

## What this does

The server knows both, so it now sends both:

- `0x0725` at login carries the saved specialization instead of zero.
- A new `ASC_LOCAL_CAD` addon message carries the specialization and the committed ranks. Same
  transport as the existing Runemaster Echoes bridge — an addon-channel whisper from the
  character to itself — chunked, because one chat packet carries a bounded payload, and applied
  only once every chunk has arrived, so a dropped packet reads as "no data" rather than as a
  short list that would look like missing ranks.

It also corrects three opcode identities. The client's opcode table in `Extensions.dll` is an
array of `mov eax, <name>; ret` stubs indexed by a pointer array, where the opcode id is the
0-based index. Decoding it reproduces every opcode in this realm's packet log exactly:

- `0x0523` is `CMSG_CUSTOM_ASCENSION_POINT_SPEND_REQUEST`, the Character Advancement point
  purchase — not a vanity delivery. It was being forwarded to the collection service, which
  would have parsed a talent purchase as an appearance request. Latent only because the shim
  overrides the native purchase path, so it has never actually been sent here.
- `0x0725`'s name was invented; the opcode is real and handled.
- `0x061A` is `CMSG_CREATURE_QUERY_BULK`.

## Scope and limits

- Ranks are still derived from the spellbook **on the server**, so this fixes the client and
  server disagreeing about a rank, not a rank the server itself got wrong. A real rank ledger
  slots in behind the same message.
- Unguarded specialization switching — clicking another spec wipes the previous one with no
  cost or confirmation — is **not** addressed here.
- The second dword of `0x0725` has an unestablished meaning and is sent as zero.

## Client side

Needs the matching patch-B, shipped in `CoA-Client-Patch.zip`: a new
`FrameXML/LocalCharacterAdvancementBridge.lua` that receives the snapshot, plus the
compatibility layer now trusting it and no longer clearing the saved specialization. Against an
older server the client falls back to the previous local behaviour.

## Overlap with #4027

#4027 ("make the server the authority on CoA talent state") covers the same area and goes
further: point-budget enforcement, `0x0726`/`0x0727` so the client's native `IsKnownID` /
`GetTalentRankByID` answer from server state, and `CharacterAdvancementEssence.dbc` for the
point curve. Two things from it correct this PR: `0x0725` carries `u32 slotIndex, u32 slotCount`
rather than a ChrSpecs id, and it has to be sent on the first `CMSG_SET_ACTIVE_MOVER` rather
than at login, because the client keys that state off a player object that does not exist yet
when `OnPlayerLogin` runs.

What is here and not there: the opcode identity corrections (`0x0523`, `0x061A`), and an
in-game verification — this branch was run against the real client today and the reported
symptoms are gone, where #4027 states its packets are not client-tested. Treat this as an
interim if #4027 is not ready to land; otherwise take the opcode-identity commit from here and
the state work from there.

Also established while investigating, and useful for #4027: `0x0725`'s first arrival is what
allocates the client's Character Advancement state object (without it `IsPending()` raises
"pending build is not available" and every specialization paints Disabled), and the client
loads its own entry catalogue from `area-52/patch-D.MPQ`, so no catalogue push is needed.

## Testing

Built (RelWithDebInfo) and run on the local realm. Verified in game: the tree opens directly
instead of the chooser, survives a logout and relog, and previously-learned talents read as
spent. `apps/codestyle/codestyle-cpp.py` passes.

Protocol notes and the decoded opcode table are in `.agents/plans/coa-cad-protocol/`
(gitignored, not part of this change).


lostmind84:
Folded into #4027 as agreed: your opcode-identity commit is cherry-picked there as its own commit (your authorship kept), and the `ASC_LOCAL_CAD` bridge is re-implemented on top of the authority code with the same message format — sent at login and again after every talent change — so your patch-B client half applies unchanged. `0x0725` stays as #4027 sends it (`(0, 1)` on the first `CMSG_SET_ACTIVE_MOVER`); your client half does not depend on it, but that combination still needs one run on a real client. Ghost e2e covers the bridge message now. Suggest closing this one in favour of #4027.

PR #4031, opened 2026-09-18
fix(CoA/Compat): persist talent builds so a reset or switch cannot lose them

bozo-1:
The realm keeps no record of a Character Advancement build. A pick's only copy is its spell in the spellbook, and a hidden (`SPELL_ATTR0_HIDDEN_CLIENTSIDE`) talent does not survive that round trip — so a specialization switch, or any save from a client that cannot read those ranks back, destroys the build for good. Nothing is stored, so nothing can be restored. In play that is the archetype chooser reopening on every login and talents reverting to a build the player never chose (#3971, #3672).

**A stored record.** The build is kept per character, keyed per tree (`core.ascension_build.<spec>`, `core.ascension_class_tree`), as `entryId * 10 + rank`, behind a marker so a tree the player emptied stays empty. A record that exists and is empty is a decision, not an absence — that is what stops the spellbook being re-read as a build after a reset.

**Only deliberate edits change it.** `.localtalent` now writes the one entry that changed instead of re-harvesting the whole spellbook, which is what folded automatic entries and left-over spells into the "preset nobody chose". An explicit `rank 0` is obeyed again: the old code refused the first one and asked for a second request, so every Reset Trees and every pick after it was silently discarded.

**Restored on login and on entering a specialization** — learn-only, so it can never take a spell away; a switch records the build it is leaving before the spellbook is touched.

**The client is told the state.** The advancement handshake answers with the stored slot and specialization instead of `0, 0`, and the same state is pushed over the addon channel the client already listens on (`ASC_LOCAL_SPEC`, `ASC_LOCAL_TALENTS`, `ASC_LOCAL_RECORDS`). The client's own assertion of a specialization it has no build for is questioned rather than obeyed, and `.localspecstate` lets the client ask for the state once its add-on has loaded, which is later than login. New config: `AscensionCompat.AnnounceActiveSpec`.

**Companion client patch required.** This repo ships the realm, not the client, so the reader/writer half lives in the client patch: two members of `patch-B` (`CharacterAdvancementCompat.lua`, `CharacterAdvancementStateCompat.lua`) that consume the messages above and send only the player's own edits. Server and client were tested together in game on a 21-class realm, and the client's send rule is covered by a Lua harness across ten class configurations. The tab-alias part of that patch (five classes drawing an empty second tree) is independent of this PR.

Applies cleanly to current `main` (1–2 line offset from #4025). It overlaps **#4027** and **#4030**, which rewrite the same advancement functions, so those and this should not be merged in either order without reconciling.


lostmind84:
Reconciled into #4027 rather than merged in either order: it now has your stored builds per specialization (written when a switch leaves a specialization, restored through the same purchase rules when one is entered, class tree shared — Ghost-tested across 56→55, relog, 56, 55), your three-message state form (`ASC_LOCAL_SPEC/RECORDS/TALENTS`) and `.localspecstate`, so your patch-B client half works unchanged against it. Not taken, with reasons in #4027's description: `0x0725` as `(slot, specId)` (measured as `(slotIndex, slotCount)`), the state send from `CanPacketReceiveEarly` (network thread), the refuse-once `.localspec` guard (a mistaken switch is reversible now), and `RefuseContradictingUnlearn`/`HasStoredBuild`/`_refusedUnlearns`, which nothing calls. Your client half is the more complete one I have seen (edits-only sends, whole adoption of the realm's ranks, `CanJudgeSpecID`); one of the two message forms should retire with whichever patch-B ships. Suggest closing this one in favour of #4027.

PR #4118, opened 2026-09-19
feat(Core): implement generic Action Bar "Activation" (Glow Effect) and wire up Chronomancer

### LINKED ISSUE #262: Spell/Ability Action Bar Glow Effect
Ascension had a feature (backported from Cata?) where ability buttons would glow. The only one I'm personally familiar with is Time Chronomancer, when you gained 5 stacks of `Endless Sands` from casting `Epoch` (meaning `Reverse Wounds` was now Instant Cast), `Reverse Wounds` would glow on your action bar.

This is similar to what a lot of people do with WeakAurus, but native.

# Action Bar Spell Activation Glow (Reverse Wound & Endless Sands)

## 1. Overview & Mechanics

When a Time Chronomancer with the  with the **Endless Sands** talent (Entry `6700`, Spell `806338`) reaches **5** stacks of **Sands of Time** (`804488`) and casts **Epoch** (Spell `801270`), the player gains the **Endless Sands** proc buff (Spell `806728`).

- **Spell 806728:** Applies `SPELL_AURA_ADD_FLAT_MODIFIER` (`SPELLMOD_CASTING_TIME`, -100%) targeting `Reverse Wound` (Rank 1: `801303`, up to Rank 9: `572628`).
- 
- **Goal:** Notify the client to activate the native golden spinning border/glow on `Reverse Wound` action buttons whenever this instant-cast condition is active, and remove the glow when consumed or expired.

---

## 2. Client Architecture (`Extensions.dll`)

Ascension's 3.3.5 client backports Cataclysm's `SpellActivationOverlayFrame` and action button overlay glow system inside `Extensions.dll` and standard FrameXML:

- **Show Event:** `SPELL_ACTIVATION_SHOW` (format: `"%u%s%s%f%u%u%u"` -> `spellID`, `texture`, `positions`, `scale`, `r`, `g`, `b`).
- **Hide Event:** `SPELL_ACTIVATION_HIDE` (format: `"%u%s"` -> `spellID`, `texture`).
- **Action Button Integration:** FrameXML's `ActionButton.lua` listens for `SPELL_ACTIVATION_SHOW` / `SPELL_ACTIVATION_OVERLAY_GLOW_SHOW` and invokes `ActionButton_ShowOverlayGlow(self)` on any action bar slot bound to matching spell IDs or spell names.

### Server Opcodes

| Opcode | Hex | Payload | Description |
|---|---|---|---|
| `SMSG_SPELL_ACTIVATION_OVERLAY_SHOW` | `0x09B1` (2481) | `uint32 spellId`, `string texture`, `uint32 locationFlags`, `uint8 scale` | Shows overlay / action button glow |
| `SMSG_SPELL_ACTIVATION_OVERLAY_HIDE` | `0x09B2` (2482) | `uint32 spellId` | Removes overlay / action button glow |

> **Note:** If `texture` is passed as an empty string (`""` or single null terminator `uint8(0)`), the client skips full-screen activation arcs and exclusively shows the action button border glow.

---

### Server-Side Opcode Dispatch (Native Server Parity)

Add a helper to dispatch the packet directly to the player:

```cpp
void SendSpellActivationGlow(Player* player, uint32 spellId, bool show)
{
    if (show)
    {
        WorldPacket data(0x09B1, 4 + 1 + 4 + 1);
        data << uint32(spellId);
        data << uint8(0); // empty string: button border glow only, no screen overlay
        data << uint32(0); // position flags
        data << uint8(0); // scale / alpha
        player->SendDirectMessage(&data);
    }
    else
    {
        WorldPacket data(0x09B2, 4);
        data << uint32(spellId);
        player->SendDirectMessage(&data);
    }
}
```

#### Aura Hook in `AscensionChronomancerTime.cpp`:
Hook the application and removal of proc spell `806728` (`Endless Sands`):
- **`OnApply`:** Call `SendSpellActivationGlow(player, 801303, true)`. (Base spell ID `801303` or the player's learned rank of Reverse Wound).
- **`OnRemove`:** Call `SendSpellActivationGlow(player, 801303, false)`.

---

I'm not sure if there's a way to recover the other abilities with this feature, but once we solve it server-side, we can implement them as people report them.

mattparkes:
Partially fixes #262. 

- Added generic SendSpellActivationGlow, SendSpellActivationShow, and SendSpellActivationHide methods to Player
- Created aura_ascension_chronomancer_endless_sands script to trigger glow based on 5 stacks
- Bound script to Endless Sands (806728) in spell_script_names
- Fixed bug where Endless Sands aura wasn't being consumed on Reverse Wound cast

Tested with a clean character and the glow and hide opcodes are sent and processed correctly.

I have generated a 1406 CSV line file from SpellActivationOverlay.dbc which should show the rest that need to be fixed for all the other specs.


modules/mod-ascension-compat/src/AscensionChronomancerTime.cpp
```diff
@@ -334,9 +339,34 @@ class chronomancer_time_contracts : public GlobalScript
...
+
+    void HandleEffectRemove(AuraEffect const*, AuraEffectHandleModes)
+    {
+        if (Player* player = GetUnitOwner()->ToPlayer())
+            player->SendSpellActivationGlow(ReverseWound, false);
+    }
+
+    void Register() override
+    {
+        OnEffectApply += AuraEffectApplyFn(aura_ascension_chronomancer_endless_sands::HandleEffectApply, EFFECT_0, SPELL_AURA_DUMMY, AURA_EFFECT_HANDLE_REAL_OR_REAPPLY_MASK);
```
jealous-sound: [P1] Bind the callbacks to Endless Sands' actual aura type

Spell 806728 has `SPELL_AURA_ADD_PCT_MODIFIER` (108) on `EFFECT_0`, but both callbacks subscribe to `SPELL_AURA_DUMMY`. `ApplyTimeContracts` does not change this spell's aura type, so the core reports an effect mismatch and never invokes either callback. Reaching five stacks therefore never sends the glow. Register both handlers against `SPELL_AURA_ADD_PCT_MODIFIER`.

Verified against the copied server `Spell.dbc` and the core's effect-hook matching code; the local database was offline, so live SQL overrides were not checked.

src/server/game/Entities/Player/PlayerMisc.cpp
```diff
@@ -133,6 +133,49 @@ void Player::SendAttackSwingDeadTarget()
...
+    data << uint32(spellId);
+    data << uint32(0);
+    
+    session->SendPacket(&data);
+}
+
+void Player::SendSpellActivationGlow(uint32 spellId, bool show)
+{
+    if (show)
+        SendSpellActivationShow(spellId);
```
jealous-sound: [P1] Connect the helper to the client's action-button activation system

With the canonical CoA client UI, these packets cannot activate button glow: the installed DLL handlers emit `SPELL_ACTIVATION_SHOW` / `SPELL_ACTIVATION_HIDE`, while `ActionButton.lua` listens for `SHOW_SPELL_ACTIVATION_OVERLAY` / `HIDE_SPELL_ACTIVATION_OVERLAY` and queries `C_SpellActivationOverlay.IsSpellOverlayed`. The legacy events only reach the screen-overlay renderer; an empty texture does not bridge the two systems. Use the action-button activation path or provide the corresponding client integration. Fixing the aura registration alone leaves this issue.

This finding is based on canonical UI source and the installed DLL. Installed UI archive verification remained incomplete, and no rendered client test was run; please account for any different UI version or additional bridge in the client used for testing.

jealous-sound:
Could you verify whether these findings are accurate?

mattparkes:
These findings _are_ semi-accurate, but there's an entire fuck around that has caused this to take me 3 days and burn endless amounts of tokens.

This spell activation glow feature actually works natively client-side without the opcodes, but there's a critical bug which means it's EXTREMELY inconsistent with anything in SpellActivationOverlay.dbc set to `TRIGGER_TYPE_STACKS` rather than `TRIGGER_TYPE_AURA_EXISTS`. Gemini was useless, but Astra found it quite quickly:

> I found a client bug and confirmed its effect in live memory by having you cast Epoch and watching the live memory values.
> `Extensions.dll` reads __two__ bytes for the aura’s stack count. The actual stack count occupies __one__ byte.
> For your Endless Sands aura, I captured:
> Actual stacks: `5`
> Adjacent byte: `65`
> Value used by glow evaluator: `16,645`
>
> The native overlay requires stacks == 5, so it fails.
>
> This explains the discrepancies:
> - Warrior of Dawn works because TRIGGER_TYPE_AURA_EXISTS never reads the stack count.
> - Endless Sands sometimes works when the adjacent byte happens to be zero.
> - Aeons may disturb it by changing aura allocation or slot reuse. That connection remains an inference; the incorrect stack read is confirmed.
> - Our server-based Opcode workaround works because it enables the overlay at five real stacks (and uses >= 5).
> The proper generic correction is in `Extensions.dll` change the overlay evaluator’s stack read from word to byte:
> ```
> movzx eax, word ptr [eax + 0xE]  ; current
> movzx eax, byte ptr [eax + 0xE]  ; correct
> ```

> The same bug exists in `Extensions.dll.ORIG`. The modifications by JS did not introduce it.
> Both files contain this instruction at file offset 0x31EBA5:


I don't think we want to patch `Extensions.dll` to fix this behaviour, so we'll likely just need to wire up the opcode-based approach for any interactions like this. 

There are 277  `TRIGGER_TYPE_STACKS` and `==` entries in the DBC file and 117 `>=` ones, but a lot of those are garbage, e.g:

```
794,0,,,0,,,TRIGGER_TYPE_STACKS,5,==,LeftRight,,1.0,1.0,1.0,1.0,0,0

```
which is totally invalid.

```
749,560170,Geomolding,Proc,681119,Terrasurge,Rank 1,TRIGGER_TYPE_STACKS,20,==,LeftRight,,1.0,1.0,1.0,1.0,0,0
```
This ==20 check will never trigger because Geomolding stacks were capped to 10 with a balance change.

A lot of the rest seem to be Mystic Enchants which I'm unfamiliar with since I only played CoA.


My current fix is to use opcode `0x6f6` to update the SpellActivationOverlay table:

> That constant is defined as 0x6F6 in Opcodes.h. The packet contains a 56-byte row followed by two NUL-terminated strings:
> - Enable: row 700, aura 806728, Reverse Wound’s spell ID, TRIGGER_TYPE_STACKS, comparison >= 5.
> - Disable: the same row with TRIGGER_TYPE_NONE.
> - The texture string is empty, so this rule requests no screen graphic.
> On the client, Extensions.dll receives it at handler RVA 0x1E3430 and adds or replaces the rule in memory. It does not modify the DBC file on disk.
> The native evaluator subsequently checks that rule. When its active state changes, it emits SHOW_SPELL_ACTIVATION_OVERLAY or HIDE_SPELL_ACTIVATION_OVERLAY; action buttons query ??C_SpellActivationOverlay.IsSpellOverlayed() and update their glow.
> So 0x6F6 changes the rule that controls the glow. The client still evaluates that rule; the packet itself is not a direct “draw glow now” command.

But we have to abuse this by sending ">= 5" when it's five and then disabling the rule when it's any other value.

I wanted to do that at login, but then we'll _still_ run into the 2 bytes vs 1 byte bug.


I'll wire up the generic feature of updating the table though, so we have that up our sleeve, and then commit this fix (which I've tested in-game with default bars and Ascensions Bartender 4 version).

mattparkes:
@jealous-sound this is now fixed and ready to be merged.

PR #4128, opened 2026-09-19
feat(Core): Render talent trees from native server packets + Dual Spec

CoinThrow:
## Changes Proposed:
This PR proposes changes to:
-  [x] Core (units, players, creatures, game systems).
-  [ ] Scripts (bosses, spell scripts, creature scripts).
-  [x] Database (SAI, creatures, etc).

> [!WARNING]  
> COLLIDES WITH CUSTOM PATCH-B.MPQ EDITS TO TALENT SYSTEM LUA FILES.

Server-side only: the original Ascension client renders its Character Advancement UI with no client or Lua changes and no memory patches.

This fixes ingame custom LUA functions related to the Realm, Talents and Spells as it's now native to Ascension using the right opcodes.

Realm cluster at character select: on the character-list request the module sends `SMSG 0x09D0 + 0x09BC + 0x06E5` before the core's `SMSG_CHAR_ENUM`. `0x09BC` carries the realm parameters whose gate bytes the client latches into its feature-flag store; with them set, the CA tree view renders and the rest of the custom pipeline initializes (verified live in client memory). The realm name comes from the new `AscensionCompat.RealmName` conf option.

Reset credits are now keyed per specialization: `core.ascension_credits.<specId>` player settings, mirroring `core.ascension_build.<specId>`. Giving up a paid rank raises the active spec's `TalentUnlearn` counter; a full `.localtalent reset` raises its `TalentReset`. The active spec's four values are sent at login and with every known-entries resend, so each spec shows its own reset prices.

Also included: the `SMSG_COA_CONFIG` data table, the known-entries record layout (asserted in the talent-state harness) and a module db-world update for spec-swap spells. Documentation: `.agents/docs/systems/ascension-talents.md` updated; 

### AI-assisted Pull Requests

> [!IMPORTANT]
> Using AI tools to prepare pull requests is allowed, but it must be disclosed and it must follow our AC guidelines for AI Agentic Engineering (link below).
>
> You are expected to fully understand the changes you submit and to be able to explain and justify them when maintainers ask.

- [x] AI tools (e.g. Claude, ChatGPT, or similar) were used entirely or partially to prepare this pull request. If checked, specify which tools and models below.
   - Tools/models used: GitHub Copilot agent (DeepSeek V4.1 Flash). The code changes and this description were prepared with AI assistance under the author's direction.
- [x] I have read and understood the [AC guidelines for AI Agentic Engineering](https://www.azerothcore.org/wiki/agentic-engineering)

## Issues Addressed:
- Closes 

## SOURCE:
The changes have been validated through:
- [ ] Live research (checked on live servers, e.g Classic WotLK, Retail, etc.)
- [x] Sniffs (remember to share them with the open source community!)
- [ ] Video evidence, knowledge databases or other public sources (e.g forums, Wowhead, etc.)
- [ ] The changes promoted by this pull request come partially or entirely from another project (cherry-pick). **Cherry-picks must be committed using the proper --author tag in order to be accepted, thus crediting the original authors, unless otherwise unable to be found**

The wire formats and feature-gate behavior were derived from the original Ascension 3.3.5a client by binary analysis and confirmed live in client memory (realm object filled, gates set, trees render with zero patches).

## Tests Performed:
This PR has been:
- [x] Tested in-game by the author.
- [ ] Tested in-game by other community members/someone else other than the author/has been live on production servers.
- [ ] This pull request requires further testing and may have edge cases to be tested.

Built and run on Windows (VS2022) against the unmodified Ascension 3.3.5a client: the Character Advancement talent tree renders and works.

## How to Test the Changes:

1. Build and run a server with `mod-ascension-compat` enabled, then connect with an Ascension 3.3.5a client — no client patches, no Lua changes.
2. Log in, enter the world and open the Character Advancement window: the tree renders with the character's known nodes.
3. On the active spec, unlearn a paid rank (or run `.localtalent reset`): that spec's reset/unlearn price goes up.
4. Switch specialization: each spec shows its own prices; switching back keeps the raised price.

## Known Issues and TODO List:
- [ ] Reset costs are only displayed through the client's cost formulas — no gold is charged yet (billing is not implemented).

## How to Test AzerothCore PRs
 
When a PR is ready to be tested, it will be marked as **[WAITING TO BE TESTED]**.

You can help by testing PRs and writing your feedback here on the PR's page on GitHub. Follow the instructions here:

http://www.azerothcore.org/wiki/How-to-test-a-PR

**REMEMBER**: when testing a PR that changes something **generic** (i.e. a part of code that handles more than one specific thing), the tester should not only check that the PR does its job (e.g. fixing spell XXX) but **especially** check that the PR does not cause any regression (i.e. introducing new bugs).

**For example**: if a PR fixes spell X by changing a part of code that handles spells X, Y, and Z, we should not only test X, but **we should test Y and Z as well**.


modules/mod-ascension-compat/src/AscensionCompat.cpp
```diff
@@ -6766,6 +7306,62 @@ class AscensionCompatAllCreatureScript : public AllCreatureScript {
...
+                AscensionClassService::Instance().SwitchSpecSlot(player, slot);
+                return;
+            }
+    }
+
+    void Register() override
+    {
+        OnCheckCast += SpellCheckCastFn(spell_ascension_spec_swap::HandleCheckCast);
+        BeforeCast += SpellCastFn(spell_ascension_spec_swap::HandleBeforeCast);
+        AfterCast += SpellCastFn(spell_ascension_spec_swap::HandleAfterCast);
```
jealous-sound, line 7364: [P1] Suppress the built-in dual-spec effect before switching CoA slots

The installed Spell.dbc gives these swap spells effect 162 (SPELL_EFFECT_TALENT_SPEC_SELECT): 979994 activates core spec 1, and 979995 requests core spec 2. That default effect runs before AfterCast, so Player::ActivateSpec has already changed core specialization, action-bar and glyph state before SwitchSpecSlot saves the outgoing CoA build. For a character with GetSpecsCount() == 2, the existing `spec > GetSpecsCount()` guard also accepts index 2 and reaches the two-element glyph array out of bounds. Suppress the default effect for these CoA swap spells and perform the slot transition explicitly; an AfterCast hook alone does not turn the original spell into a marker.

modules/mod-ascension-compat/src/AscensionCompat.cpp
```diff
@@ -1807,6 +2105,63 @@ class AscensionClassService {
...
+        return;
+      requests = std::move(itr->second);
+      _pendingInspects.erase(itr);
+    }
+
+    if (!IsAscensionCustomClass(player))
+      return;
+    for (uint64 const requested : requests)
+    {
+      Player* target = ObjectAccessor::FindPlayer(ObjectGuid(requested));
```
jealous-sound: [P1] Keep inspection on the target player's owning map thread

ProcessInspectRequests runs from the viewer's OnPlayerUpdate, but ObjectAccessor::FindPlayer can return a player on another map. SendCharacterAdvancementInspectResult then reads that player's spellbook and settings while their own map can update them. ActiveSpecSlot/StoredSlotSpec also use GetPlayerSetting, which inserts or resizes missing settings, so this is not even a read-only cross-thread access. With parallel map updates, these accesses race spell and settings changes and can corrupt state or crash. Use a same-map lookup with the appropriate inspection checks, or collect an immutable snapshot on the target's owning thread before answering.

modules/mod-ascension-compat/src/AscensionCompat.cpp
```diff
@@ -1252,10 +1299,18 @@ class AscensionClassService {
         _activeSpecializations[player->GetGUID().GetCounter()] = specializationId;
     }
 
+    // Slot one has no tome: the client's specification list gates every row on its swap spell being
+    // known, and Tomes of Specialization only exist from slot two on. Grant it so the starting slot
+    // stays reachable, like on a live character.
+    if (!player->HasSpell(ASCENSION_SPEC_SWAP_SPELLS[0]))
+        player->addSpell(ASCENSION_SPEC_SWAP_SPELLS[0], SPEC_MASK_ALL, true);
```
jealous-sound: [P2] Notify the client when granting the first-slot spell

OnPlayerLogin runs after SendInitialPacketsBeforeAddToMap has already sent the initial spell list. This permanent addSpell call updates server state but does not send SMSG_LEARNED_SPELL; that notification is sent by learnSpell. The native CASpecListMixin gates each slot with IsSpellKnown, so a character receiving 979993 for the first time still sees Specialization I as unavailable until another login and cannot select it after switching away. Grant it through learnSpell or explicitly send the learned-spell notification.

modules/mod-ascension-compat/src/AscensionCompat.cpp
```diff
@@ -1943,6 +2298,181 @@ class AscensionClassService {
...
+    if (currentSlot == targetSlot)
+    {
+      SendCharacterAdvancementState(player);
+      return true;
+    }
+
+    uint32 currentSpec = GetActiveSpecialization(player);
+    if (!currentSpec)
+      currentSpec = StoredSlotSpec(player, currentSlot);
+    if (currentSpec)
```
jealous-sound: [P2] Preserve class-tree builds before a specialization is selected

A character can purchase class-tree entries while GetActiveSpecialization returns 0. In that state this condition skips StoreSlotBuild, but the following loop still removes every class talent spell. Returning to that slot also skips RestoreSlotBuild because targetSpec is 0, permanently losing those class-tree choices. Store and restore the slot's class-tree picks even when its specialization ID is zero; the absence of a chosen specialization does not imply an empty build.

modules/mod-ascension-compat/src/AscensionCompat.cpp
```diff
@@ -1486,6 +1681,99 @@ class AscensionClassService {
...
+  /// slot with its stored picks over the automatic entries the slot's specialization always carries.
+  std::vector<AscensionCoATalentState::KnownEntry> SlotKnownEntries(Player* player, uint32 slot)
+  {
+    if (slot == ActiveSpecSlot(player))
+      return KnownTalentEntries(player);
+
+    uint32 const slotSpec = StoredSlotSpec(player, slot);
+    std::vector<AscensionCoATalentState::KnownEntry> known;
+
+    for (AscensionCoATalentState::KnownEntry const& live : KnownTalentEntries(player))
```
jealous-sound: [P2] Reconstruct inactive automatic talents from the inspected slot

For an inactive slot with a different specialization, KnownTalentEntries(player) contains the active specialization's spellbook. Switching slots has already removed the inactive specialization's automatic grants, and LiveSlotPicks deliberately excludes automatic entries from the saved picks. Consequently neither this loop nor the stored-build loop includes those grants, so the inspection response presents an incomplete inactive build. Derive automatic entries from the inspected slot's specialization, level and stored prerequisites instead of filtering the currently active spellbook.

lostmind84:
I went through `Extensions.dll` (the stock one, `Extensions.dll.ORIGINAL`, SHA-256 `0f8d847b…`) with Ghidra and a lab client, mostly to verify this PR rather than to review it. Some of it confirms your work, some contradicts it. Everything below is reproducible from the binary.

**Two of your addresses re-derived independently.** The client registers handlers as `push <handler>; push <opcode>; call <register>`; scanning `.text` for that shape gives 532 registrations, and opcode `0x09BC` lands on handler RVA `0x002fc6c0` — your `0x102FC6C0` at the image base. Likewise the Lua binding table (built as consecutive `{const char* name, void* fn}` writes) puts `ApplyPendingBuild` at RVA `0x001742c0`, your `0x101742C0`. Neither was reached by looking at your figures.

**Opcode names, from the client's own table.** It has a switch (`cmp eax,0x9d4; jmp [eax*4+0x102c7af0]`) whose jump table maps every opcode to its name string — 2517 entries, 2059 named. It confirms `0x06E5 = SMSG_CHAT_INFRACTION`, `0x0926 = SMSG_CA_AVAILABLE_CREDITS_UPDATE`, `0x09BC = SMSG_REALM_INFO`, and it also confirms #4030's `0x0523`, `0x061A` and `0x064A` exactly.

Two corrections from the same table:

- `0x09D0` is `SMSG_CHARACTER_CUSTOMIZATION_UNLOCKS`, not a pending-selection flush. Worth re-checking what it actually does before describing it as a reset.
- `0x0900` has **no name and no handler registration** in this build, and neither does `0x0903 SMSG_REPLICATE_CONFIG`. If that is right, `SMSG_COA_CONFIG` sent on `0x0900` cannot be doing anything client-side, which also makes it an unlikely candidate for the input `ApplyPendingBuild`'s validator is missing.

**On `IsKnownID` requiring `[+0x10] == 1` and a rank at `[+0x14]`.** Both offsets are real and in the function you cite, but that is a hash-map node (FNV-1a over the entry id, sentinel at `[service+0x234]`), and `[+0x14]` is used as an **array index** into `[service+0x1a8]` and `[service+0x1b4]` — the `0 / 0x2c` test is a range check, not a rank test.

**The `0x0726` record.** The deserializer (`0x00166760`, your `0x00166899` is its per-record loop) reads `u32, u32, u32, u8, u32, u32` into a `0x20`-byte record at `+0x04, +0x08, +0x0c, +0x10, +0x18, +0x1c`. Measured against a running client by forging packets with AzerothCore's own `.debug send opcode`:

- presence of a record is what makes `IsKnownID` true — the `u8` is not required for that;
- the `u8` is a **lock flag**: `0` leaves the entry unlocked, `1`, `2` and `255` all make `IsLockedID` true while `IsKnownID` stays true. A locked entry is one the UI won't let you interact with, which looks like "the entry didn't work" without being the same thing. That may be what you saw in game;
- `marker` at `0`, `1`, `2` changed nothing observable.

**Why ranks never show, and why it isn't a server problem.** Three accessors read one vector (`[service+0x20]`, stride `0x20`, key `+0x04`): membership, the rank at `+0x08`, the lock byte at `+0x10`. So the rank field is right where the module already puts it.

`C_CharacterAdvancement.GetTalentRankByID` consults the **Wildcard** store first (`0x2c` records, filled by `SMSG_WILDCARD_ENTRY_LEARNED 0x0621` — its handler `0x00a30260` is the function that push_backs into it). Outside Wildcard mode that store is empty, so it falls back to the talent store keyed on `*(catalogue_entry + 0)`, which is not the entry id and never matches.

One forged record proves it: `u8 = 1` made `IsLockedID` answer true, which requires `+0x04 == 31194` in that vector, while `+0x08 = 5` on the same record still read back as `0`. Same vector, same key field, different key. The native UI reads ranks only through that getter (`SpellListItem.lua:79`, `CharacterAdvancementUtil.lua:71`, `CharacterAdvancementCostUtil.lua:138`, `CASpellButton.lua` ×3, `CharacterAdvancement.lua:932`), so on this build no server can make it display a talent rank.

**The fix is a client-side one.** The binding passes `*(entry + 0)` where it should pass the entry id. That is a small in-place edit of the kind the CoA client patch already makes — it currently changes 30 bytes across ten sites in `.text` plus an appended `.local` section, and moves no code, so RVAs stay valid. Feeding the Wildcard store instead would work but would collide with the actual Wildcard game mode, so it isn't a real option.

Practical consequence for this PR: the compat shim isn't a stopgap for a missing server feature, it's working around a broken client binding, and it can't retire until that binding is patched.

Happy to share the extraction scripts (opcode table, handler map, Lua bindings — all plain Python over the PE, no disassembler needed for the first two). One thing that would help a lot: the SHA-256 of the `Extensions.dll` your addresses came from. Two matched here, but knowing it's the same build would let the rest be trusted directly instead of re-derived.

Issue #4091, opened 2026-09-18
[WD/SH] Spirit Volley not working properly


coldcite:
(Related to #1506)

There are a few issues with Spirit Volley talent (503712) the level 50 keystone ability from the Witch Doctor's Shadowhunting talent tree.

**Expected (as it worked in CoA)**
- Spirit Volley buff (505158) procs on you
- Reclamation (572345) is transformed into Spirit Volley ability (504582)
  _NOTE: While the buff was active, I could still cast Reclamation if I used it with /macro_
- Cast your new Spirit Volley ability (504582)
- Target is hit
- On hit (or miss?) Spirit Volley buff (505158) will fade off you
  _NOTE: This way, if you were quick, you could fire 2 volleys every proc_

**What's happening here**
- Spirit Volley buff (505158) procs on you
- Reclamation (572345) is transformed into Spirit Volley ability (504582)
  However, in this case, Reclamation completely disappears from you, ie. you CAN NOT cast it with /macro 
- Ability learned spam on chat 
  <img width="415" height="174" alt="Image" src="https://github.com/user-attachments/assets/725de0cc-923c-46dc-9062-b59df88918cb" />
- Cast your volley
- On cast end, Spirit Volley buff (505158) fades off you
- Spirit Volley ability (504582) transforms back into Reclamation (572345)
- Ability learned spam on chat again, plus toast
  <img width="498" height="198" alt="Image" src="https://github.com/user-attachments/assets/b3dd0446-63a7-4b10-801d-5cb328a617ac" />
  <img width="443" height="505" alt="Image" src="https://github.com/user-attachments/assets/2026f224-8d68-4b8c-8dc3-8c1bb5d7eed8" />

**Some side notes**
- I'm not exactly sure when the buff was granted back on CoA, if on hit or on cast complete. Here's is granted on cast complete
- Don't ask me how, but for some reason, If the buff timed out while you were still casting, you could still complete that cast somehow, here you will stop casting the volley as soon as the buff fades off you.
- Unlike #1506, I'm pretty sure the whole being able to cast Reclamation while Spirit Volley was active was intentional, since at one point they nerfed Spirit Voley damage so "you could choose to use Reclamation on specific scenarios and not just Spirit Volley whenever is available" as per their patch notes.
- Just like with #1506, the icon sometimes bugs out.

Vanheden:
[...]

Issue #1506, opened 2026-09-16
[WD/SH] Umbral Glaive not working properly


coldcite:
There are a few issues with Umbral Glaive talent (681242) the level 30 keystone ability from the Witch Doctor's Shadowhunting talent tree.

**Expected (as it worked in CoA)**
- Umbrail Glaive buff (504588) procs on you
- Hex of Malice (501153) is transformed into Umbral Glaive ability (501134)
 _NOTE: While the buff was active, I could still cast an Hex of Malice if I used it with /macro_
- Cast your new Umbral Glaive ability (this ability is off the GCD)
- Target is affected with the Umbral Galive debuff (501134)
- After a small delay, DoT/leech starts
_NOTE: As far as I can remember applying a new debuff refreshed its duration, but did not restart the DoT timer_
- Your Umbrail Glaive buff (504588) is consumed
- After that, Umbral Glaive turns back into Hex of Malice

**What's happening here**
- Umbrail Glaive buff (504588) procs on you
- Hex of Malice (501153) is transformed into Umbral Glaive ability (501134) 
 _However, in this case, Hex of Malice completely disappears from you, ie. you CAN NOT cast it with /macro_
- Ability learned spam on chat
  <img width="529" height="106" alt="Image" src="https://github.com/user-attachments/assets/6268c673-dce1-4960-aa27-84a0d121d866" />
- Fire your Umbrail Glaive ability
- Target is affected with the Umbral Galive debuff (501134)
- DoT starts ticking _right away_
- You Umbrail Glaive buff (504588) IS NOT consumed
- If you also triggered you Mojo Madness talent (705781) it WILL NOT be consumed either
- You can constantly fire the Glaive over and over, dealing tons of damage (ability off the GCD + free + DoT/leech ticks on application)
- Umbrail Glaive buff (504588) runs out after 10 seconds
- Umbrail Glaive ability transforms back into Hex of Malice 
- Ability learned spam on chat again, plus toast
  <img width="505" height="104" alt="Image" src="https://github.com/user-attachments/assets/7aea1a81-1858-457a-bbfc-132646ac7283" />
  <img width="402" height="95" alt="Image" src="https://github.com/user-attachments/assets/861bd087-9a36-43b1-ae81-8bf063b154c4" />

**Some side notes**
- Glaive projectile had a faint shadow aura VFX around it, but that's just a minor VFX detail, I actually kinda like it better without.
[...]

PR #4253, opened 2026-09-19
fix(CoA/WitchDoctor): Umbral Glaive buff, Spirit Volley cast, Spirit Eclipse spirits

### LINKED ISSUE #273: [WD] Spirit Eclipse consumes your spirits (high)
Even though Spirit Eclipse (801607) description indicates "Unleash your Spirits on an enemy" so it might lead to believe it would, this ability never "consumed" the Spirits on you.
The only ability that consumed Spirits was "Frenzied Spirits" (560748) at the bottom of the tree, as indicated by its description.

Vanheden:
## Changes Proposed

Three Witch Doctor fixes, grouped because the two proc-buff ones (#1506, #4091) share a root cause in the WD cast hooks and #273 sits in the same code.

- **Umbral Glaive never consumed its buff (#1506).** Instant spells are cast from inside `Spell::prepare`, before the prepare hook runs, so the "which proc buffs were up" snapshot was empty and the consume step did nothing. The snapshot is now also taken at the first cast check. The glaive's own ranged hit could also re-proc the Umbral Glaive talent and refill the buff it had just spent, so it no longer counts as a proc source.
- **Spirit Volley cast stopping when the buff runs out (#4091).** The cast check re-runs when the cast completes and failed once the buff had expired. A cast that began under the buff now completes.
- **Spirit Eclipse consumed Spirits (#273).** It now leaves them alone; only Frenzied Spirits consumes them. A Price To Pay and Dambala's Lethality still trigger off Eclipse.

Fixes #273

Refs #1506 and #4091: they are not fully resolved. Still open on those two issues: keeping the base spell (Hex of Malice / Reclamation) castable by macro while the replacement is active, the "ability learned" chat spam and toast, the icon flicker, the DoT start delay and timer on refresh, and off-GCD / multi-target mana cost on the glaive. The macro and spam items depend on how the client handles `SMSG_SUPERCEDED_SPELL`, which I could not observe.

- [x] Core / Scripts (mod-ascension-compat)

## AI Disclosure

Changes and this description were written by Claude Code (Sonnet 5).

## Tests Performed

New gameplay scenarios, run on a build of this change:
- `witchdoctor-umbral-glaive.json`: failed before (buff not consumed), passes after; also asserts Hex stays known and the mana charge is not multiplied by the extra targets.
- `witchdoctor-spirit-volley.json`: failed before (cast aborted when the buff was removed 100 ms in), passes after; a normal cast consumes the buff.
- `witchdoctor-spirit-eclipse-keeps-spirits.json`: Eclipse damages the enemy and the 3 Spirits remain; Frenzied Spirits then consumes them. Not run against the old build.
- Not tested in-game (client behaviour above).

## How to Test

As a Witch Doctor with Umbral Glaive, get the proc, cast Umbral Glaive once and check the buff is gone. With Spirit Volley, let the buff expire during a Volley cast and check it still lands. Gather Spirits, cast Spirit Eclipse and check they are still there.

## How to Test AzerothCore PRs

Follow https://www.azerothcore.org/wiki/how-to-test-a-pull-request


jealous-sound (approved):

Vanheden:
Follow-up commit for #1506: Umbral Glaive's leech now first ticks one period after the hit, re-applying keeps the periodic timer (core: PERIODIC_LEECH treated like a DoT on reapply), and the glaive is off the GCD. Scenario `witchdoctor-umbral-glaive` passes on a build of this branch. Not done: casting base Hex by macro while the buff is active (the core swaps the base ID via the temporary replacement, so the server cannot tell), and the learn-spam/toast/icon items, which depend on client handling of SMSG_SUPERCEDED_SPELL. Off-GCD and timer-preservation were not directly asserted; nothing tested in-game. Issue stays `Refs #1506`.

PR #4280, opened 2026-09-19
fix(CoA): revive the Reaper's dead talents and abilities

Mrpirolito:
Reaper talents and abilities that were learnable but did nothing. Everything here is the class itself; no tier set, no items.

### Talents whose proc never fired

A talent written as `SPELL_AURA_PROC_TRIGGER_SPELL` still needs a `spell_proc` row, because the spell record carries no proc flags of its own. Seventeen Reaper talents had none, so their aura sat on the character and nothing ever triggered. Each row here names the event the tooltip describes: a melee ability landing, an attack being dodged or parried, a periodic tick, damage taken.

Fourteen more were the same shape but keyed to an ability list rather than a family mask. `AscensionReaperTalentProcs.{h,cpp}` holds those fourteen rules as `{talent, spells}` pairs behind one shared `AuraScript`, so a talent that only applies to, say, the seven ranks of Slaughter checks the list instead of a mask that cannot express it.

Sixteen talents had a `SpellClassMask` of all zeroes, which matches nothing. They are filled from what the tooltip says the talent modifies.

### Extinction

`spell_ascension_reaper_extinction` gives the talent its rolled chance, `5 + 10 * reapedSoulStacks`, and `spell_ascension_reaper_extinction_buff` narrows the consumption to the seven ranks of Slaughter so an unrelated cast cannot eat the buff.

### Ruin

The proc row was enough for the damaging rank 801322 cast on its own, and not enough for the way the ability is used. The button casts the transform 572382, whose aura triggers 801322 every 100 ms, and a spell cast by a periodic aura tick does not reach the proc system the way a cast does, so the mark never landed in play. Rather than widen the proc flags until something sticks, 801322 applies the mark itself when its caster owns the talent. One place, both routes.

### Redshade

The talent's proc applies Thresh (Dummy) 525058, whose tooltip says Reap has transformed into Thresh, and casting Thresh then leaves Bloodshatter (Dummy) 525299 for the second step. Both are plain dummies as far as the server is concerned, so all three parts of the promise are the module's: the cast has to be legal, the button has to say what the cast will do, and neither can announce itself in chat. Each part ruled out the obvious shortcut.

**The cast.** `CMSG_CAST_SPELL` is refused outright for a spell the caster does not have active — "cheater? kick? ban?" — so a button that merely looks like Thresh spends a global cooldown and does nothing. A Reaper never learns Thresh or Bloodshatter, because they exist only as what Redshade turns Reap into, so owning the talent is what puts them in the spellbook, through `aura_ascension_reaper_redshade_spells` on Redshade 524735, for as long as the talent is taken.

**The silence.** `Player::learnSpell` answers a temporary learn with `SMSG_LEARNED_SPELL`, which the client prints as "You have learned a new ability: Thresh" — once per transform, with the spec synchronisation the learn hooks re-ran announcing every rank of Reap behind it. `Player::addSpell` with `learnFromSkill` set is the one way in that sends no packet at all. `SendInitialSpells` is no way around it either: it prints one line per ability the client did not already have.

**The button.** `SMSG_ACTION_BUTTONS`, the packet a druid's forms already use to swap a whole bar. The client is handed the character's own bar with every button holding a rank of Reap reading as the replacement, and gets it back unchanged when the last step falls off; nothing is written, so the saved buttons stay as the player left them. `Player::SetTemporarySpellReplacement` sends `SMSG_SUPERCEDED_SPELL` instead, which prints that same learned-ability line for whichever spell it names — including the rank of Reap it hands the button back to — so it announced the transform twice per cast. Ascension aura 337 on the two dummies does not move the button on its own.

`spell_ascension_reaper_redshade_reap` covers the other direction: a client that still sends the rank on the bar gets the replacement cast for it. It is bound to the eight ranks the button casts and not to 801327, which is named Reap as well but whose rank reads "Heal" — a self-targeted heal that has nothing to do with the strike.

Both transformed abilities also read "Generates 1 Soul Fragment" and "generating Runic Power", and neither record performs either, so the transformed button paid nothing while the ability it replaced paid in full. They get the two `ResourceGainRules` rows every rank of Reap already has.

**Known, and left as it is:** Bloodshatter's buff is not consumed when its cast turns the button back into Reap, so the second step can be used more than once inside its ten seconds. The buff is a plain dummy with no charge data of its own, and the tooltip does not say which cast is meant to spend it.

### Haunt

Haunt summoned creature 840000, the Witch Doctor's Mirage, whose script opens with a class check and returns for anyone else. A Reaper therefore got a neutral level one creature that stood where it was cast: no copied appearance, no faction, no owner, no movement.

990020 is the Reaper's own visage under `npc_ascension_reaper_haunt`. It copies the caster's appearance through Clone Me! 45204 (`SetDisplayId` alone gives the untextured base race model), takes their faction and level so enemies can hit it, takes their GUID as owner, and runs 40 yards in a random direction rather than the one the caster happens to face. What keeps the caster out of the fight is that the visage never fights back: passive, no combat movement, an empty `AttackStart`.

### Haunt's charges, and every other Reaper cooldown

Haunt handed its charges straight back. They were being spent — the stored pool read two of three — but its recovery time read one millisecond instead of thirty seconds, so the charge returned before it could be noticed as gone.

`SpellInfo::IsAffected` reads an empty `EffectSpellClassMask` as "every spell of the family". That is right for the stock classes and wrong for these: an Ascension talent names the abilities it changes and carries their mask, so a row without one is unfinished data, not a wildcard. Four Reaper talents carry a cooldown modifier with no mask, worth -30000, -10000, -5000 and -4000 milliseconds, and each was applying to every cooldown and every charge recharge the class has. Their sum lands past zero, and the clamp in `ConsumeSpellCharge` turns that into the one millisecond floor.

Where the tooltip names an ability that has a mask of its own the mask is filled in, in the client data. Two name abilities that carry no mask at all — Crimson Scythe and Spectral Waltz — so no mask can express them, and `DisarmUnmaskedCooldownModifier` disarms the modifier rather than leave it pointing at the whole class. It is limited to `SPELLMOD_COOLDOWN` and `SPELLMOD_GLOBAL_COOLDOWN` on families 18-38, so nothing else changes shape.

### Soulstone Lure

Four separate problems. The summon used display 408534, a model this client does not hold, so it rendered as a chequerboard; it is now 9832, which the world already spawns. It followed the player, because `SummonProperties` 61 is Control 1, a pet; 64 is the stationary row the ability wants. It came out at ziggurat size, and is now 0.12, the width of the ring the ability draws. And the model's origin sits at its middle, so at any scale it was half buried: ground movement `Hover` is what `CanHover()` reads and what puts `MOVEMENTFLAG_HOVER` on the creature, and `HoverHeight` is how far off the ground it then sits.

## Tests

`apps/coa-gameplay-test` covers the new behaviour, as asked in review:

- `reaper-ruin-shudder-scythe.json` — the damaging rank marks its target, and the mark expires.
- `reaper-redshade-thresh.json` — the talent owns both transformed abilities, the transformed cast lands and pays a Soul Fragment and Runic Power, the second step follows it, Bloodshatter can then be cast in its own right, and the saved action button is still Reap while all of it happens.
- `reaper-haunt-visage.json` — the visage is the caster's own summon, the cast spends a charge, the spent charge is still gone inside the recharge window, and the caster stays out of combat.

All three pass against a worldserver run by `apps/coa-gameplay-test/run.py`.

Three additions to the harness say what these needed. `spell_charges` reads `Player::GetSpellCharges`, because charges are a pool separate from cooldowns and a charged spell reports no cooldown at all until the pool empties. `set_action_button` and `action_button` put a spell on the bar and read back what the saved button holds. The Haunt scenario also enters Underwalk 800797 first, because Haunt's `ShapeshiftMask` allows only Stealth and Shadowform.


jealous-sound:
Could you please cover new behaviour with coa-gameplay-test?

Mrpirolito:
> Could you please cover new behaviour with coa-gameplay-test?

Done, everything is working, posted imgs and message in the reapers thread

jealous-sound (approved):

Issue #1966, opened 2026-09-17
Crafting - No XP and single skill up


jealous-sound:
Submitted from the in-game bug-report form.

#### Issue
Leveling First Aid is not giving any XP and you only get 1 skill increase at a time. In live CoA you got 2 skill up per, and a small amount of XP (sorry, don't know exactly how much, maybe half a mob kill?)

### Classification
Category: 9
Severity: 0

Client: 3.3.5 (12340)

### Server context
Class ID: 20
Level: 15
Map ID: 1
Position: -2356.5, -353.634, -8.95668
Core revision: be725043c415

lostmind84:
Confirmed: crafting gives no experience; that logic is missing server-side (lostmind84/ConquestOfAzerothGhost#6).
One skill point per craft is the server default setting (SkillGain.Crafting = 1). How much experience should a craft give?

PR #4289, opened 2026-09-20
feat(CoA): grant character experience for gathering and crafting

### LINKED ISSUE #368: Crafting/Gatheing no longer gives character XP

#### Issue
In ascension crafing or gathering used to give normal character xp, this functionality seems to have been entirely server-side and as such is not present here.
No
#### What were you doing
Gathered a herb
#### Expected Outcome
Some small 20-50xp gain
#### Actual Outcome
No xp gain, just profession proficiency

### Classification
Category: 9
Severity: 0

Client: 3.3.5 (12340)

### Server context
Class ID: 20
Level: 6
Map ID: 1
Position: 9701.67, 808.773, 1293.17
Core revision: 7f0761f621fc

### LINKED ISSUE #1410: No XP From Crafting

#### Spell
 Any Profession [ID: Not Included]

#### Issue
The original CoA had crafting which gave experience, this is missing from this version at this time. Example: Crafting an orange level bandage would give the same XP as killing a monster of your own level.
No
#### Exact Location
N/A
#### What were you doing
Crafting
#### Expected Outcome
Extra XP from crafting
#### Actual Outcome
None Gained
#### Steps to Reproduce
Craft any item before max level

### Classification
Category: 0
Severity: 0

Client: 3.3.5 (12340)

### Server context
Class ID: 16
Level: 13
Map ID: 530
Position: -4135.42, -11764.8, -136.835
Core revision: 2abd8e291ade

### LINKED ISSUE #4266: Profession skill ups do not grant experience

#### Issue
Player experience should be received upon crafting items that produce a profession skill up.
No
#### Exact Location
Anywhere
#### What were you doing
Crafting first aid.
#### Expected Outcome
Skill ups would increase character level experience.
#### Actual Outcome
No experience granted upon profession skill up.

### Classification
Category: 9
Severity: 0

Client: 3.3.5 (12340)

### Server context
Class ID: 23
Level: 33
Map ID: 0
Position: 1509.96, 154.762, -62.17
Core revision: 7f7131c76ac3

elChivoR:
The original CoA paid ordinary character experience for mining, herbing, skinning and crafting, and that was lost in the move to this core. Four reports ask for it back: #368, #1410, #1966 and #4266.

A new module, `mod-profession-xp`. No core changes.

## How much

The award is what a mob of the player's own level is worth, taken from `Acore::XP::BaseGain` so it follows the core's own curve, scaled by the colour of the node or recipe against the player's skill — orange, yellow, green, and grey which pays nothing. The core already computes those thresholds for its skill-up roll and hands them to the hook, so nothing is recomputed here.

No canonical formula survived, and the reports disagree on the amount:

| Report | What it remembers |
| --- | --- |
| #1410 | an orange craft paid *"the same XP as killing a monster of your own level"* |
| #1966 | *"a small amount... maybe half a mob kill?"* |
| #368 | *"some small 20-50xp gain"* from a herb at level 6 — this grants 60 there at the full rate, so about half |

Two of the three describe half a kill, so `balanced` (0.50 orange) is the default. `realpath` (1.00) matches #1410, and `CraftingFactor` lets a realm pay half for gathering and full for crafting, which satisfies all three at once. The choice is going to a realm vote; every value is re-read on `.reload config`, so settling it needs no rebuild.

## Why colour is the only guard

Colour alone stops a node being farmed, because skill only ever rises: an orange node becomes yellow, then green, then grey, and stops paying.

An earlier revision also scaled the award down by the node's tier against the player's level. It was removed after in-game testing, because the two measures disagree — colour is relative to skill and tier to level. A level 22 who had just learned mining was shown an **orange** copper vein that the tier rule called trivial and paid **nothing** for, which reads as the feature being broken.

## When

On every gather and craft, not only the ones that raise the skill. #368 reports experience for gathering a herb, which frequently does not skill up. That covers #4266 as well, since a craft producing a skill up is paid like any other.

## Following the player's XP rate

`Player::GiveXP` does not raise `OnPlayerGiveXP` itself — each caller in the core raises it first. The module does the same, so this experience obeys the rate a player chose with `.xp` instead of always arriving at the realm's base rate. Without it, a player on `.xp 5` would earn five times as much from a mob as from a node.

## Recipes without trivial data

Six First Aid entries (3273, 3274, 7924, 10846, 27028, 45542) carry `TrivialSkillLineRankHigh = 0`, which this module reads as grey and pays nothing for. Those are the profession rank spells (Apprentice/Journeyman/…), not craftable bandages — the bandages all carry proper data (Linen Bandage: high 60, low 30). `SkillGainChance` with `grey = 0` returns the grey chance, so the core grants no skill-up for them either; the behaviour matches.

## Validation

- Builds clean on MSVC Release, no warnings.
- `apps/codestyle/codestyle-cpp.py` passes on both new files.
- Deployed to a local repack: the module and its conf load, no config errors.
- In-game: gathering was tested and produced the copper bug described above, which is what prompted removing the tier scaling. **The corrected build's numbers have not yet been confirmed in-game** — a level 22 mining an orange node should award 78 XP on the default preset. Crafting has not been tested in-game at all.

## Not included

#1966 has a second half — *"you only get 1 skill increase at a time, in live CoA you got 2"*. That is `SkillGain.Crafting` in `worldserver.conf`, currently `1`, so it is a realm config change rather than code. Hence `Refs #1966` and not `Fixes`.

Fishing is also not covered: `Player::UpdateFishingSkill` uses a probability ladder instead of the colour thresholds and would need a rule of its own.

Fixes #368
Fixes #1410
Fixes #4266
Refs #1966


jealous-sound:
We have it seems original config values for exact xp gain for looting resources and stuff. Give me some time, I'll add this original config values, so we can reuse them here.

elChivoR:
Closing as superseded by #4670, which adds profession XP in core with per-profession rates, fishing/disenchanting coverage, and a dedicated XP source that also resolves the NO_EXPERIENCE_EXCEPT_PROFESSIONS clash in mod-coa-challenges. Running both would award profession XP twice.

Issue #4489, opened 2026-09-21
Cultist Heratic - Level 10 passive Herald of the Depths 92131 does not teach the cultist Herald of the Depths spell


jealous-sound:
Submitted from the in-game bug-report form.

#### Spell
 Herald of the Depths [ID: 92131]

#### Issue
Level 10 passive does not automatically teach the player Herald of the Depths spell. It is unlearnable.

Comment: It may be worthwhile checking other specs for the same issue.
#### Is this a Gamebreaking Issue
No
#### Exact Location
Anywhere
#### What were you doing
Playing heratic cultist
#### Expected Outcome
Automatically learn Herald of the Depths when selecting Heratic cultist spec
#### Actual Outcome
Does not learn spell
#### Steps to Reproduce
Reach level 10 cultist
Select heratic spec

### Classification
Category: 1
Severity: 0

Client: 3.3.5 (12340)

### Server context
Class ID: 25
Level: 27
Map ID: 1
Position: 9856.05, 1572.88, 1313.62
Core revision: 83831e0d3b06

artysmith1:
Spell ID: 520326 (Herald of the Depths)

PR #4491, opened 2026-09-21
fix(CoA): Stormbringer, Cultist, Bloodmage, Barbarian fixes + Docker module shipping, coa-world mysql fix

### LINKED ISSUE #4448: Stormbringer - Altered Course Not Functioning

#### Spell
 Altered Course [ID: 707053]

#### Issue
This talent does not work, I press the discharge ability with it talented but I do not get the movement speed increase.
No
#### Exact Location
N/A
#### What were you doing
N/A
#### Expected Outcome
Speed increase after casting Discharge.
#### Actual Outcome
No speed increase.
#### Steps to Reproduce
Talent Altered Course as a Stormbringer and cast Discharge, you won't get the movement speed increase.

### Classification
Category: 1
Severity: 2

Client: 3.3.5 (12340)

### Server context
Class ID: 16
Level: 27
Map ID: 0
Position: -10952.8, -1375.3, 58.4602
Core revision: b3717c137710+

Deathresist:
## Summary
- **fix(coa-world)**: drop `--no-login-paths` from the mysql client invocation - that flag isn't recognized by every mysql client build in use, causing the query helper to fail before it can connect.
- **feat(Docker)**: ship `modules/` into the worldserver image and activate module confs - the image built its binary against `modules/` but never copied the directory itself in, so module data/config not compiled into the binary was missing at runtime.
- **fix(CoA/Stormbringer)**: Altered Course's speed buff never fired on Discharge (#4448). Altered Course (707053) is a zero-proc-flag Proc Trigger Spell dummy, so its trigger spell (707219, the movement-speed buff) never fires natively; Discharge's own cast hook now supplies that cast event, resolved through the spell's rank chain so it still matches after Discharge auto-upgrades past level 10.
- **fix(CoA/Cultist)**: Herald of the Depths (92131), the Heretic spec's level-10 identity passive, was never granted (#4489). `AscensionCompat.AutoProgression` defaults off and isn't set on most servers, and spec switches synchronize progression without the override that bypasses it. Identity passives are meant to bypass that toggle per ChrSpecs' own design ("a Level 10 Passive that must not wait for a purchased class ability"), so they're now flagged and granted regardless of the toggle, while everything else it gates is unaffected. Also fixed: casting Insanity was never reset to 0 when the Herald transform buff expired, despite that being in its own tooltip.
- **fix(CoA/Stormbringer)**: Torrential Wrath (Tempest Sovereign's Call Lightning replacement) never consumed Static, and once it did, used Call Lightning's 50 cost instead of its own 40 (#4494). Resource rules are authored against a talent's original spell, but a talent-replaced spell's cast packet carries one of its own replacement rank IDs, which never matched any rule. Added a resolver applied everywhere the same unresolved-ID pattern existed (cast, hit, periodic-tick, and pre-cast validation hooks) - it affected every class with a talent-replaced spell, not just this one - plus an explicit 40-Static cost entry for Torrential Wrath's own rank IDs, checked before the fallback to Call Lightning's rule, so a replacement with its own authored cost is not shadowed by the ability it replaces.
- **fix(CoA/Bloodmage)**: Blood Rituals never healed on Bloodmoon Blast cast, despite Dark Liturgy correctly applying the buff and its own one-time AOE heal (#4505). The recurring chip heal is a separate chain - 706623 should trigger 704118 ("Blood Rituals Heal Proc Source"), a Proc Trigger Spell dummy pointing at 704119 (a self-centered AOE heal) - again authored with 100% proc chance but zero proc flags, so it never fires natively. Bloodmoon Blast's own cast hook now supplies that cast event when the caster has the Blood Rituals buff.
- **fix(CoA/Cultist)**: Hallucination's three appearance copies rendered as a blank white model instead of the caster's own (#4498). The clone's native `creature_template_model` used `CreatureDisplayID` 49, an unused placeholder that renders solid white with no character texture. It also relied on `player->CastSpell(me, 49889, true)` to apply the mirror-image clone aura, but the aura never landed on the summon (`HasCloneCasterAura()` stayed false), so the client's mirror-image data request went unanswered and the clone rendered invisible once flagged. Fixed the native model fallback and grant `SPELL_AURA_CLONE_CASTER` directly with `player->AddAura(49889, me)` so display and equipment resolve from the caster as intended.
- **fix(CoA/Barbarian)**: Spirit of Savagery appeared to never activate, and Spirit of the Hunt had the same problem (#4511). Two separate issues: (1) the "only one Spirit stance active at a time" check resolved the caster via `Owner(GetTarget())`, but these stances radiate their real effect onto nearby enemies (`SPELL_EFFECT_APPLY_AREA_AURA_ENEMY`), so `GetTarget()` there is always the affected enemy, never the caster - the exclusivity check was dead code, letting two Spirits sit active on the same enemy at once. Fixed by resolving the caster via `GetCaster()` instead. (2) Spirit of Savagery and Spirit of the Hunt specifically carried `SPELL_ATTR0_AURA_IS_DEBUFF`, which forces every effect on the spell negative, including their harmless self-only marker effect (unrelated to the real enemy debuff, which stays correctly negative on its own via its dedicated `APPLY_AREA_AURA_ENEMY` rule). `SpellMgr` bakes that verdict into `AttributesCu`'s cached `NEGATIVE_EFFn` bits at load time, so clearing the raw attribute alone wasn't enough - the cached bit needed clearing too (same pattern already used in `AscensionXorothContracts.cpp` for an identical situation). With the self effect forced negative, the client never recognized the caster as carrying the aura, so the shapeshift-bar stance button never lit up the way Spirit of Conflict/War/Bloodshed (which never carried this attribute) already did. No gameplay-data changes - the enemy-facing debuff mechanic was already correct by design per each spell's own tooltip text.
- **fix(CoA/Reaper)**: Doomrend was dealing damage and applying its healing-absorb shield but never generating Runic Power, despite its own tooltip promising it (#4517). Its `Spell.dbc` record has no energize effect at all - only school damage, weapon-percent damage, and a trigger to the healing-absorb helper (560361) - and no `spell_proc`/`spell_bonus_data`/`spell_custom_attr` override fills the gap either, so nothing ever granted the power. Added it to the existing `NativePowerGainRules` table (already used for this exact "tooltip promises a gain the DBC doesn't implement" situation) at 150 internal tenths (15 Runic Power) on a successful hostile hit, matching the placeholder value already used for Soulrend, which carries the same untooltipped "generates Runic Power" wording.

Fixes #4448, #4489, #4494, #4505, #4498, #4511, #4517.

## Test plan
- Altered Course: rebuilt and confirmed in-game the speed buff now applies after Discharge.
- Herald of the Depths: rebuilt and confirmed in-game the passive is granted, Total Madness is correctly suppressed at 100 Insanity, and Insanity resets to 0 when the transform buff expires.
- Torrential Wrath: rebuilt and confirmed in-game it now consumes exactly 40 Static.
- Blood Rituals: rebuilt and confirmed in-game casting Bloodmoon Blast while the buff is active now heals the caster and nearby allies.
- Hallucination: rebuilt and confirmed in-game the three copies now show the caster's own character model and gear.
- Barbarian Spirits: rebuilt and confirmed in-game that Spirit of Savagery's enemy debuff lands correctly, that activating a second Spirit correctly removes the first from an affected enemy, and that the stance-bar button now glows for Spirit of Savagery and Spirit of the Hunt.
- Doomrend: confirmed in-game that Runic Power is now generated on hit, alongside the existing damage and healing-absorb effects.


jealous-sound:
No gameplay test scenarios was covered, use "coa-gameplay-test" in your agent for details

PR #4514, opened 2026-09-21
fix(CoA/Reaper): show the spirit world while in Underwalk

anonymoose1671:

Underwalk (800797) reads "Sneak around in the spirit world, becoming stealthed". The Reaper stealths and is slowed, but the screen never changes: no desaturation, no spirit world sky, no ghost ambience.

800797 itself only carries the stealth shapeshift, the stealth and the movement slow. The look is authored in a hidden companion spell, 414054 "Reaper Stealth Ghost Visual": a single `SPELL_AURA_SCREEN_EFFECT` with MiscValue 1 (ScreenEffect 1, the ghost presentation). Nothing on the server applied it.

This adds one `spell_linked_spell` row (`SPELL_LINK_AURA`, type 2) so 414054 is applied with Underwalk and removed with it, whichever way Underwalk starts or ends. The core needs no handler for `SPELL_AURA_SCREEN_EFFECT`; the client resolves the MiscValue against its own ScreenEffect.dbc once the player holds the aura.

- `python apps/codestyle/codestyle-sql.py --files <the new file>`: all checks passed.
- 414054 and 800797 were read from the repack's server `Spell.dbc`: 414054 exists there with the effect described above, draws no buff icon and is not passive, so it is sent to the client.
- Not built and not run against this repository's server. The same link (800797 -> 414054, type 2) has been running on the contributor's own private server, where the spirit world effect shows and clears correctly; that server needed extra `spell_dbc` rows which this one does not, since it loads the original CoA Spell.dbc.
- 414054 keeps its authored 20 minute duration while Underwalk is permanent, so a Reaper who stays in Underwalk for more than 20 minutes without a break loses the visual until they re-enter it.

## Data/source dependencies

Spell data from the original CoA `Spell.dbc` already loaded by the server. No new spells, no client changes.

Confirm that applied SQL is unchanged and that no credentials, profiles or game archives are included: confirmed, the only change is one new file in `data/sql/updates/pending_db_world/`.


anonymoose1671:
I noticed the spirit world effect was missing when I tried reaper on the PTR so I figured I'd try my hand at a simple fix. Tested locally and appeared to work and disable upon leaving stealth properly.

